PR #1, opened 2016-08-18
WIP init data

hzoo:
No logic at all yet just getting something out there.


baer:
Can you merge this as is (or at least some kind of no-op version of it) and I'l start working on working on it a bit? I'd LOVE to help out on this. 

I actually began writing a similar project nearly 2 years ago that came at this from the other angle. I was using tern.js to do code inference and lint based on a browser matrix. it got a bit out of hand and never got finished but needless to say, I am very excited about this kind of work. https://github.com/caniuse-js/caniuse-js

hzoo:
sure might as well

PR #2, opened 2016-08-31
Warm people in the README that this is a wip project

baer:


hzoo:
It's not published either, but ok

PR #3, opened 2016-09-01
Implement very basic (but working) functionality

baer:
This is a VERY basic first pass at getting this plugin working. This plugin should more or less work for the following cases:

* If no options are provided, all plugins will be applied
* If options are provided they must be provided as numbers. E.g `chrome: 55`
* If loose-mode is specified, the name must match what is in the plugin file
* If modules are specified, they must be an Array of module types.

Obviously there is plenty to do to make this robust and working for all cases especially around getting all of the data idiosyncrasies right. I didn't want to go much further than this though without a PR since big PRs are not only hard to review but I wanted a 👍 on the direction I'm headed.

In addition to the functionality, the following changes were made

* Plugins are defined in a separate file
* Mocha was added for some very simple testing
* Plugins where environments had features that had been implemented from the beginning were changed from positive infinity to negative infinity to make the comparison easier.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,9 +54,7 @@
     "babel-plugin-transform-object-rest-spread": "^6.8.0",
     "babel-plugin-transform-proto-to-assign": "^6.9.0",
     "babel-plugin-transform-regenerator": "^6.6.0",
-    "lodash.filter": "^4.6.0",
-    "lodash.includes": "^4.3.0",
-    "lodash.map": "^4.6.0"
+    "lodash.includes": "^4.3.0"
   },
   "devDependencies": {
     "babel-cli": "^6.11.4",
--- a/package.json
+++ b/package.json
@@ -13,18 +13,16 @@
     "lint": "eslint src test",
     "fix": "eslint src test --fix",
     "ci": "npm run lint && npm run test",
-    "prepublish": "npm run build"
+    "prepublish": "npm run build",
+    "test": "mocha ./test"
   },
   "dependencies": {
-    "babel-plugin-transform-es2015-modules-amd": "^6.8.0",
-    "babel-plugin-transform-es2015-modules-commonjs": "^6.6.0",
-    "babel-plugin-transform-es2015-modules-systemjs": "^6.12.0",
-    "babel-plugin-transform-es2015-modules-umd": "^6.12.0",
-    "babel-plugin-transform-es3-member-expression-literals": "^6.3.13",
-    "babel-plugin-transform-es3-property-literals": "^6.3.13",
-    "babel-plugin-transform-proto-to-assign": "^6.9.0",
-    "babel-plugin-transform-es5-property-mutators": "^6.3.13",
     "babel-plugin-check-es2015-constants": "^6.3.13",
+    "babel-plugin-syntax-trailing-function-commas": "^6.13.0",
+    "babel-plugin-transform-async-to-generator": "^6.8.0",
+    "babel-plugin-transform-class-properties": "^6.11.5",
+    "babel-plugin-transform-decorators-legacy": "^1.3.4",
+    "babel-plugin-transform-do-expressions": "^6.8.0",
     "babel-plugin-transform-es2015-arrow-functions": "^6.3.13",
     "babel-plugin-transform-es2015-block-scoped-functions": "^6.3.13",
     "babel-plugin-transform-es2015-block-scoping": "^6.6.0",
@@ -35,6 +33,10 @@
     "babel-plugin-transform-es2015-for-of": "^6.6.0",
     "babel-plugin-transform-es2015-function-name": "^6.3.13",
     "babel-plugin-transform-es2015-literals": "^6.3.13",
+    "babel-plugin-transform-es2015-modules-amd": "^6.8.0",
+    "babel-plugin-transform-es2015-modules-commonjs": "^6.6.0",
+    "babel-plugin-transform-es2015-modules-systemjs": "^6.12.0",
+    "babel-plugin-transform-es2015-modules-umd": "^6.12.0",
     "babel-plugin-transform-es2015-object-super": "^6.3.13",
     "babel-plugin-transform-es2015-parameters": "^6.6.0",
     "babel-plugin-transform-es2015-shorthand-properties": "^6.3.13",
@@ -43,16 +45,18 @@
     "babel-plugin-transform-es2015-template-literals": "^6.6.0",
     "babel-plugin-transform-es2015-typeof-symbol": "^6.6.0",
     "babel-plugin-transform-es2015-unicode-regex": "^6.3.13",
-    "babel-plugin-transform-regenerator": "^6.6.0",
+    "babel-plugin-transform-es3-member-expression-literals": "^6.3.13",
+    "babel-plugin-transform-es3-property-literals": "^6.3.13",
+    "babel-plugin-transform-es5-property-mutators": "^6.3.13",
     "babel-plugin-transform-exponentiation-operator": "^6.8.0",
-    "babel-plugin-syntax-trailing-function-commas": "^6.13.0",
-    "babel-plugin-transform-async-to-generator": "^6.8.0",
-    "babel-plugin-transform-class-properties": "^6.11.5",
-    "babel-plugin-transform-decorators-legacy": "^1.3.4",
-    "babel-plugin-transform-do-expressions": "^6.8.0",
     "babel-plugin-transform-export-extensions": "^6.8.0",
     "babel-plugin-transform-function-bind": "^6.8.0",
-    "babel-plugin-transform-object-rest-spread": "^6.8.0"
+    "babel-plugin-transform-object-rest-spread": "^6.8.0",
+    "babel-plugin-transform-proto-to-assign": "^6.9.0",
+    "babel-plugin-transform-regenerator": "^6.6.0",
+    "lodash.filter": "^4.6.0",
+    "lodash.includes": "^4.3.0",
+    "lodash.map": "^4.6.0"
   },
   "devDependencies": {
     "babel-cli": "^6.11.4",
@@ -61,10 +65,19 @@
     "eslint": "^3.3.1",
     "eslint-config-babel": "^1.0.1",
     "eslint-plugin-babel": "^3.3.0",
-    "eslint-plugin-flow-vars": "^0.5.0"
+    "eslint-plugin-flow-vars": "^0.5.0",
+    "mocha": "^3.0.2"
   },
   "babel": {
     "plugins": [
+      "babel-plugin-transform-es2015-destructuring",
+      "babel-plugin-transform-es2015-function-name",
+      "babel-plugin-transform-es2015-modules-commonjs",
+      "babel-plugin-transform-es2015-parameters",
+      "babel-plugin-transform-es2015-shorthand-properties",
+      "babel-plugin-transform-es2015-spread",
+      "babel-plugin-transform-es2015-sticky-regex",
+      "babel-plugin-transform-es2015-unicode-regex",
       "transform-class-properties",
       "transform-flow-strip-types"
     ]
--- a/src/index.js
+++ b/src/index.js
@@ -4,172 +4,14 @@
 // "proto-to-assign",
 // "es5-property-mutators",
 
-export const pluginsList = {
-  // es2015
-  "transform-es2015-arrow-functions": {
-    edge: 13,
-    firefox: 45,
-    chrome: 49,
-    opera: 36,
-  },
-  "transform-es2015-block-scoped-functions": {
-    edge: 11,
-    firefox: 46,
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -13,18 +13,16 @@
     "lint": "eslint src test",
     "fix": "eslint src test --fix",
     "ci": "npm run lint && npm run test",
-    "prepublish": "npm run build"
+    "prepublish": "npm run build",
+    "test": "mocha ./test"
   },
   "dependencies": {
-    "babel-plugin-transform-es2015-modules-amd": "^6.8.0",
```
baer: This was just a sort from running `npm install --save-dev mocha`

package.json
```diff
@@ -61,10 +65,19 @@
     "eslint": "^3.3.1",
     "eslint-config-babel": "^1.0.1",
     "eslint-plugin-babel": "^3.3.0",
-    "eslint-plugin-flow-vars": "^0.5.0"
+    "eslint-plugin-flow-vars": "^0.5.0",
+    "mocha": "^3.0.2"
   },
   "babel": {
     "plugins": [
+      "babel-plugin-transform-es2015-destructuring",
```
baer: I ❤️ writing in ES-next and this is the basic set of plugins missing from Node 4.x. If you'd prefer I dev with Node 6 and take this out or some other thing, I'm happy to do so.

src/index.js
```diff
@@ -190,12 +32,71 @@ export const stagePlugins = [
...
+  const modulesMode = getModules(opts.modules);
+  const targets = getTargets(opts.targets);
 
-  if (typeof stage !== "boolean") throw new Error("Preset es2015 'loose' option must be a boolean.");
+  const transformations = Object.keys(pluginList)
+    .filter(pluginName => isPluginRequired(targets, pluginList[pluginName]))
+    .map(pluginName => {
+      return includes(looseMode, pluginName)
+        ? [require(`babel-plugin-${pluginName}`), { loose }]
+        : require(`babel-plugin-${pluginName}`);
```
baer: I used the es2015 preset as an example here. If I don't need to do the full require or if it's okay to have an Array with just one element like `[require(`babel-plugin-${pluginName}`)]` it would clean up this statement a little bit.

test/index.js
```diff
@@ -0,0 +1,32 @@
+"use strict";
+
+const babelPresetEnv = require("../lib/index.js");
```
baer: NOTE: the tests here rely on built code. Happy to switch it up and run tests through babel or something if you'd prefer.

package.json
hzoo: Since babel supports node 0.10 still maybe we should be using the same transform as that (although we want to drop it in babel 7)? We can always make this 4.x only as well

src/index.js
```diff
@@ -190,12 +32,71 @@ export const stagePlugins = [
...
+    });
 
+  const modules = Object.keys(modulesMode)
+    .map(moduleType => {
+      return [require(`babel-plugin-transform-es2015-modules-${moduleType}`), { loose }]
+    });
+
+  return {
+    plugins: [].concat(transformations, modules)
```
hzoo: I did a true/false check in es2015 and filtered by Boolean, https://github.com/babel/babel/blob/master/packages/babel-preset-es2015/src/index.js#L66-L72

You should be able to use `...` instead of concat here I think?

src/index.js
hzoo: I think the short version is if you pass in a string.

Where is `loose` coming from? it's not referenced anywhere I think?

I don't know if we want to just pass loose to each plugin, but then you can just do `[require(`babel-plugin-${pluginName}`), { loose: includes(looseMode, pluginName) }]`

test/index.js
hzoo: Not a big deal, if we get everything working we can also just move it into babel/babel

package.json
```diff
@@ -43,16 +45,18 @@
...
-    "babel-plugin-transform-do-expressions": "^6.8.0",
     "babel-plugin-transform-export-extensions": "^6.8.0",
     "babel-plugin-transform-function-bind": "^6.8.0",
-    "babel-plugin-transform-object-rest-spread": "^6.8.0"
+    "babel-plugin-transform-object-rest-spread": "^6.8.0",
+    "babel-plugin-transform-proto-to-assign": "^6.9.0",
+    "babel-plugin-transform-regenerator": "^6.6.0",
+    "lodash.filter": "^4.6.0",
+    "lodash.includes": "^4.3.0",
+    "lodash.map": "^4.6.0"
```
hzoo: I don't think filter/map is used? and we can always just do indexOf !=== -1. We can change all of this later since we just want something working

package.json
baer: Humm, good to know. Well, this very plugin would be useful here so until we can just use `./lib` in place of this whole list lets pull transformations in as needed. Once we're comfortable with where everything is I'll go back over and pull in extra transformations for Node 0.10.

baer: https://github.com/babel/babel-preset-env/issues/4

src/index.js
baer: loose comes in as an option. So far I have:

```
{
  modules: []
  loose: []
  targets: [] 
}
```

Sorry, I didn't update docs since I didn't know if you would even want to accept this. Would you like me to keep the README up to date as I go?

baer: Ahhh, I see now that the var loose is not defined. Sorry bout that. Fix incoming.

src/plugins.js
```diff
@@ -0,0 +1,131 @@
+export default {
+  // es2015
+  "transform-es2015-arrow-functions": {
+    edge: 13,
+    firefox: 45,
+    chrome: 49,
```
Kovensky: Function.name won't work correctly, at least in some cases, with untransformed arrow functions. `const x = () => {}; x.name` may become `"(anonymous)"`, but it should be `"x"`.

src/plugins.js
```diff
@@ -0,0 +1,131 @@
...
+    chrome: 41,
+    opera: 28,
+  },
+  "transform-es2015-block-scoping": {
+    edge: 11,
+    firefox: 36,
+    chrome: 49,
+    opera: 36,
+  },
+  "transform-es2015-classes": {
```
Kovensky: This needs to be enabled wherever super and spread are not both available at the same time.

This is easier to trigger when using `transform-class-properties`, but `class X extends Y { constructor (...args) { super(...args) } }` cannot be transformed by spread, as it's not possible to call `super` as a Function object.

Going by the version numbers you specified, this seems to be ok with browsers, but, again, this is a problem in Node 4.

src/plugins.js
```diff
@@ -0,0 +1,131 @@
...
+    chrome: 44,
+    opera: 31
+  },
+  "check-es2015-constants": {
+    edge: 11,
+    firefox: 36,
+    chrome: 49,
+    opera: 36,
+  },
+  "transform-es2015-destructuring": {
```
Kovensky: This is required (through bad factoring) by `transform-object-rest-spread`. `transform-object-rest-spread` actually only transforms spread, the rest transform is done by this.

src/plugins.js
```diff
@@ -0,0 +1,131 @@
...
+    chrome: 49,
+    opera: 36,
+  },
+  "transform-es2015-block-scoped-functions": {
+    edge: 11,
+    firefox: 46,
+    chrome: 41,
+    opera: 28,
+  },
+  "transform-es2015-block-scoping": {
```
Kovensky: Some code that builds with the transform will fail at runtime with native block scoping, but in this case it's the user's fault for not respecting the TDZ.

src/plugins.js
baer: I'm not sure I follow what you're saying. Is this an issue with the transformation or with the implementation in that version of node, with the data or with this specific comparator. Can you suggest something different?

I mainly intend to follow a simplified version of the data the [caniuse DB](https://github.com/Fyrd/caniuse/blob/master/CONTRIBUTING.md#supported-changes) follows. Of course we won't care much about a partial implementation since we just need to know whether or not to apply a plugin.

baer:
@hzoo I think I've addressed all your feedback in that last commit and responded to your comments. Let me know if you'd prefer for me to rebase onto the latest master.

src/plugins.js
```diff
@@ -0,0 +1,131 @@
...
+  },
+  "transform-es2015-destructuring": {
+    edge: 14,
+    firefox: Number.NEGATIVE_INFINITY,
+    chrome: 52,
+    opera: 39,
+  },
+  "transform-es2015-for-of": {
+    edge: Number.NEGATIVE_INFINITY,
+    firefox: Number.NEGATIVE_INFINITY,
```
Kovensky: Are you sure ancient firefox's `for of` follows the ES6 spec, with `@@iterator` support and all?

src/plugins.js
baer: Oooof this is just the initial data from @hzoo. Thanks for this audit - it looks like it's going to be complicated! Maybe we should address the plugin data/relationships in follow on PRs after the core stuff is working. Thoughts?

src/plugins.js
baer: I didn't do any kind of auditing of this data for correctness but I'm impressed with all of your comments! Do you have a source for this or are these just things you've noticed over the years. Either way, it'd be great to have your help getting all the data set up correctly.

Kovensky: This may also need to be enabled on any environments where core-js would polyfill `Symbol`, assuming this is targeting use with `babel-polyfill`.

src/plugins.js
```diff
@@ -0,0 +1,131 @@
...
+    chrome: 49,
+    opera: 36
+  },
+  "transform-es2015-template-literals": {
+    edge: 13,
+    firefox: 34,
+    chrome: 41,
+    opera: 28,
+  },
+  "transform-es2015-typeof-symbol": {
```
Kovensky: (Not sure but) this may be needed on environments where core-js polyfills `Symbol`

src/plugins.js
```diff
@@ -0,0 +1,131 @@
...
+    chrome: 41,
+    opera: 28
+  },
+  "transform-es2015-object-super": {
+    edge: 13,
+    firefox: 45,
+    chrome: 49,
+    opera: 36,
+  },
+  "transform-es2015-parameters": {
```
Kovensky: I haven't checked, but this may be needed wherever destructuring is not available.

If destructuring is done in the parameter list, the destructuring transform may not see it.

src/plugins.js
Kovensky: Node implemented arrow functions, but not contextual function name, so all arrow functions end up unnamed even though it should have received a contextual name.

Basically, if you enable `transform-es2015-function-name` in an environment, you also need to enable `transform-es2015-arrow-functions`, as the function name transform can only work on function expressions. The expected result is similar to `const x = function x() {}`.

src/plugins.js
Kovensky: I opened a bug about the untransformed class + transformed spread problem (https://phabricator.babeljs.io/T7467), but no solution appeared so far.

I propose throwing a compilation error asking for `transform-es2015-classes` to be enabled if it ever encounters a `super` with spread arguments, but that's something the spread plugin has to do.

src/plugins.js
Kovensky: @baer My other comments were mostly from experience, though this one is guessing :)

I just know that `for of` has to work with any iterable (i.e. anything with a `Symbol.iterator`), and I wouldn't expect very old firefox would have an implementation supporting that, even if it works with the basic case of Array.

src/plugins.js
```diff
@@ -0,0 +1,131 @@
...
+    chrome: 49,
+    opera: 36,
+  },
+  "transform-es2015-shorthand-properties": {
+    edge: 12,
+    firefox: 33,
+    chrome: 43,
+    opera: 30
+  },
+  "transform-es2015-spread": {
```
Kovensky: Both this and destructuring are needed if core-js needs to polyfill `Symbol` (array rest/spread actually uses iterators). This is probably not a problem in practice, though; or at least any engines would have needed a working `Symbol` before implementing this.

src/plugins.js
```diff
@@ -0,0 +1,131 @@
...
+  "transform-es2015-unicode-regex": {
+    edge: 12,
+    firefox: 46,
+    chrome: 51,
+    opera: 38,
+  },
+  "transform-regenerator": {
+    edge: 13,
+    firefox: 45,
+    chrome: 49,
```
Kovensky: This might still be needed in Chrome due to https://bugs.chromium.org/p/v8/issues/detail?id=5322

It works fine with defaults, but breaks with harmony enabled, or in Canary defaults.

hzoo:
@Kovensky maybe we should just remove the data since a lot of it is probably wrong.

hzoo:
@baer yeah we just want a poc to iterate over. 👍

PR #5, opened 2016-09-02
Improve module option input and add loose mode to modules

baer:
My goal here is to get the API fully working with the test data we have (or something very similar). Once we're all happy with how it works then it's just a matter of updating the data to tag which plugins are required in which versions + whatever bits of interdependency we have to code for.

---

This PR does the following
* Enable `loose` mode for modules. It was a one liner so I don't know why I didn't just do this before
* Finish the `opts.modules` parsing function such that it can accept a boolean, a string or an array of module types.
* Tests!

I'm pretty sure that we can call the module section feature complete after this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,8 @@
     "babel-plugin-transform-object-rest-spread": "^6.8.0",
     "babel-plugin-transform-proto-to-assign": "^6.9.0",
     "babel-plugin-transform-regenerator": "^6.6.0",
-    "lodash.includes": "^4.3.0"
+    "lodash.includes": "^4.3.0",
+    "lodash.isarray": "^4.0.0"
   },
   "devDependencies": {
     "babel-cli": "^6.11.4",
--- a/src/index.js
+++ b/src/index.js
@@ -4,6 +4,7 @@
 // "proto-to-assign",
 // "es5-property-mutators",
 
+import isArray from "lodash.isarray";
 import includes from "lodash.includes";
 import pluginList from "./plugins.js";
 
@@ -12,14 +13,6 @@ export const plugins = [
   "es3-property-literals",
   "proto-to-assign",
   "es5-property-mutators",
-];
-
-// modules?
-export const modules = [
-  "transform-es2015-modules-amd",
-  "transform-es2015-modules-commonjs",
-  "transform-es2015-modules-systemjs",
-  "transform-es2015-modules-umd"
 ];
 
 export const stagePlugins = [
@@ -31,6 +24,13 @@ export const stagePlugins = [
   "transform-function-bind",
   "transform-object-rest-spread",
 ];
+
+export const MODULE_TRANSFORMATIONS = {
+  "amd": "transform-es2015-modules-amd",
+  "commonjs": "transform-es2015-modules-commonjs",
+  "systemjs": "transform-es2015-modules-systemjs",
+  "umd": "transform-es2015-modules-umd"
+};
 
 /**
  * Determine if a transformation is required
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -54,7 +54,8 @@
     "babel-plugin-transform-object-rest-spread": "^6.8.0",
     "babel-plugin-transform-proto-to-assign": "^6.9.0",
     "babel-plugin-transform-regenerator": "^6.6.0",
-    "lodash.includes": "^4.3.0"
+    "lodash.includes": "^4.3.0",
+    "lodash.isarray": "^4.0.0"
```
baer: `Array.prototype.isArray` can't come fast enough!

test/index.js
```diff
@@ -29,4 +31,48 @@ describe("babel-preset-env", () => {
...
+    describe("Boolean argument", () => {
+      it("should return an empty Array if passed a falsy value", () => {
+        assert.deepEqual(babelPresetEnv.getModuleTransformations(), []);
+        assert.deepEqual(babelPresetEnv.getModuleTransformations(null), []);
+        assert.deepEqual(babelPresetEnv.getModuleTransformations(false), []);
+      });
+      it("should return an Array of all module transformations if passed `true`", () => {
+        const allModuleTransformations = Object
+          .keys(MODULE_TRANSFORMATIONS)
+          .map(keys => MODULE_TRANSFORMATIONS[keys]);
```
baer: Also, `Object.prototype.values` will be really really nice.

package.json
hzoo: I thought Array.isArray works in node?

hzoo: Ok, we can always change all that later - I just thought it was in node 0.10

baer:
If you're okay with this approach I might do a first pass on the data so that we have proper transformations applied for 2 options:
* Node 6
* ES5 (all options)

After that we can add Node 4, then iojs then node 10 then start working on browsers via the caniuse DB. Thoughts? My guess is that the company I work for ([Formidable](https://github.com/FormidableLabs)) would happily donate some dev time to getting those data tasks done.

baer:
Actually, now that I think about it we _could_ start with all of the presets (excluding React) as a first pass which would make it immediately useful even before we figure out all of that data. Obviously I won't work towards that unless you think it's worth doing but I did see that in your initial commit you have the `stagePlugins` variable which made me think that this was something you intended to support.

screen shot 2016-09-02 at 12 43 18 pm

```json
{
  "presets": [
    ["env", [
      "es2015", "es2016", "es2017", "stage0"
    ]]
  ]
}
```

```json
{
  "presets": [
    ["env", "latest"]
  ]
}
```

```json
{
  "presets": [
    ["env", "es2015"]
  ]
}
```

hzoo:
Passing in the presets seems weird to me because you then it seems like the preset is just a fancier way of just doing `"presets": ["es2015", "es2016", "es2017", "stage0"]` (but maybe that's beneficial too)

```
{
  "presets": [
    ["env", {
      "targets":  { "chrome": 52 }
    }]
  ]
}
```

baer:
It was a proposal to expand of the scope of the project. In effect it would make something like this a one-stop-shop for setting up the compilation of standard ES language features. You'd of course pull in more plugins/presets for things like react, minification etc.

The way I was starting to think about it is this: if you want to move from the model babel has now which is "enable what the standards committee says is approved in the lang" to the more real wold "enable only what you need" then something like this could eventually take the place of things like `babel-preset-es2015`. 

The reason for including the ESXXXX plugins like I mentioned is that a lot of companies use a tiered browser matrix. For example, a co might say "we don't support IE9 but we are not going to intentionally break it." In other words, bug fixes won't be prioritized but they would still *want* to over-compile just so they don't screw those customers over intentionally. For those types of users, the blanket language feature / stage presets would still be useful.

*Customers for whom blanket compilation makes more sense would do this*
 ```json
{
  "presets": [
    ["env", [
      "es2015", "es2016", "es2017", "stage0"
    ]]
  ]
}
```

*Customers that want to keep compilation minimal and have a well defined browser matrix*
```json
{
  "presets": [
    ["env", {
      "chrome": 45,
      "firefox": 30
    }]
  ]
}
```

Anyway, it was just a thought and it would certainly be easy enough to do. Might help DRY up some things in the codebase but I don't have strong feelings either way.

hzoo:
I guess I'm just saying that 

```
{
  "presets": [
    ["env", [
      "es2015", "es2016", "es2017", "stage0"
    ]]
  ]
}
```

is just doing

```
{
  "presets": ["es2015", "es2016", "es2017", "stage0"]
}
```

baer:
You're right but it's would be possible for a future release to only include one configurable preset as opposed to a half dozen individual ones. In the meantime it's an easy way to test out getting this to work so I'll try it. We can always kill that feature later if it's contentious.

Let me know if there is any feedback for this PR specifically.

src/index.js
```diff
@@ -72,15 +72,34 @@ const getLooseMode = looseOpts => {
...
+/**
+ * Given modulesOpts return an Array of all required module transformations
+ * @param  {String|Array.<String>|Boolean}  modulesOpts  An argument enumerating the modules to transform
+ * @return {Array.<String>}  An array of module transformation plugin names
+ */
+export const getModuleTransformations = modulesOpts => {
   if (!modulesOpts) { return []; }
-  return modulesOpts;
+
+  if (modulesOpts === true) { return Object.values(MODULE_TRANSFORMATIONS); }
```
hzoo: You should only want 1 module transform or none?

baer: Is there some incompatibility between the transformations? I assumed it would be possible to turn all of them on and transform any module type it encountered. I usually do module transformations in my bundler (require.js/webpack/browserify) so it's very possible that I misunderstood how Babel does it.

hzoo: each module turns import -> whatever syntax (amd,cjs,systemjs,umd). You would only want to use one of the transforms or none (keep import). I'm just saying no one would want to choose multiple

(Although internally it does use them)

hzoo: Sorry I'm going to push a change that does what I mean above ^. It's actually just the same logic as es2015.

hzoo:
https://github.com/babel/babel-preset-env/pull/6

src/index.js
baer: No problem - I was away for the long weekend! I DID misunderstand this. I understood it to be UMD/CJS/etc => cjs and not ES2015 => UMD/CJS/etc. 

Now I know. Hope you had a great labor day!

Issue #7, opened 2016-09-06
Data issues


hzoo:
~~Data can be wrong in https://github.com/babel/babel-preset-env/blob/master/src/plugins.js.~~

We're using https://kangax.github.io/compat-table/es6/ now

<img width="1316" alt="screen shot 2016-10-08 at 11 13 43 am" src="https://cloud.githubusercontent.com/assets/588473/19214029/58deebce-8d48-11e6-9004-ee3fbcb75d8b.png">

There is also the issue of bugs in babel, bugs in the implementation, incomplete implementations (it supports the basic test of a feature but not everything), etc. This will require a lot of work to figure out but shouldn't be too bad with testing/attention to detail.

Can also use https://github.com/sdkennedy/babel-preset-target

fson:
I took a first stab at generating the data from ES compatibility table in #8.

chicoxyzzy:
Hey! Compatibility table collaborator here! I just want you to know I will be working on having raw  data in a distinct npm package soon.

hzoo:
We should figure out granularity

- some browsers implement under a flag: (compat-table has `flag, strict, etc`) - what are these?
- maybe the browser implements most of the feature (say 12/13) which covers most use cases etc.

hzoo:
we'll need to make some PRs to `compat-table` as we go when we find issues/missing stuff

DZuz14:
Any stuff a junior dev can work on? Will gladly help out with anything I can.

hzoo:
One thing that we want to do (continually) is check that `data/plugins.json` is actually correct/accurate to both what's actually true (testing in browsers/node) as well as correct to kangax.github.io/compat-table/es6/

hzoo:
We're using compat-table already so we can close this. The issue is just that the data might be not up or date or the logic incorrect but that's an ongoing effort

PR #8, opened 2016-09-18
Build the browser data from ES compatibility table

fson:
Use the data from https://github.com/kangax/compat-table to automatically build the browser data.

Each Babel plugin is mapped to a list of features in the compatibility table (in `data/pluginFeatures.js`), and the minimum implementing version looked up from the compatibility test data. The script builds the final browser data file in `data/plugins.json`.

Notes:
* Version numbers are still marked with integers to match the previous format, but we'll probably want to move to strings so we can support environments like Node 0.12 and Android 5.1.
* Opera is not included in the compatibility table as a separate browser.


hzoo:
@fson Awesome can try tol look at it this weekend! We should figure out how to merge incorporate https://github.com/sdkennedy/babel-preset-target which already seems to do this as well @sdkennedy

hzoo:
@kangax @zloirock if you want to review (or know others that would be interested)

chicoxyzzy:
@hzoo hi! I am interested in this and as I said [here](https://github.com/babel/babel-preset-env/issues/7#issuecomment-249359363) I'll be working on this in compat-table repo. But we need to move all browsers definitions to a single module first because we have some inconsistent info there. I'll start this week. Also I'll try to find time to review this so you can use this until I finish all work in compat-table.

Looking forward to collaborate with Babel team to make great things together!

hzoo:
Oh ok I didn't see that comment for some reason 👍

package.json
```diff
@@ -61,11 +62,14 @@
     "babel-plugin-transform-flow-strip-types": "^6.8.0",
     "babel-preset-es2015": "^6.14.0",
     "babel-register": "^6.14.0",
+    "compat-table": "github:kangax/compat-table#gh-pages",
```
zloirock: In this form, it's definitely unsafe. We can have breaking changes. Required https://github.com/kangax/compat-table/issues/711

sdkennedy:
Sorry for taking so long to respond. I'm pretty busy with work right now, but feel free to steal any code from my project haha.

https://github.com/sdkennedy/babel-preset-target

package.json
hzoo: Makes sense. It is still blocked by us having to run the build script to generate the json so we can at least view that beforehand

So if we don't auto run `npm run build-data`  and commit it it can be reviewed

hzoo:
@zloirock aha I got caught up on `safari51` and others like that - that could be `safari5_1` or something so will have to make a map for now (or make the logic account for that)

hzoo:
Won't make a release and will merge for now and send another PR

hzoo:
Made https://github.com/babel/babel-preset-env/pull/12 - please take a look

@sdkennedy yep sounds good 😄

PR #12, opened 2016-10-06
fixes, run build script in node 0.10 with babel-node

hzoo:
Followup of https://github.com/babel/babel-preset-env/pull/8

### DIFF
--- a/data/plugins.json
+++ b/data/plugins.json
@@ -25,7 +25,7 @@
     "chrome": 44,
     "edge": 12,
     "firefox": 34,
-    "safari": 10
+    "safari": 7
   },
   "check-es2015-constants": {
     "chrome": 49,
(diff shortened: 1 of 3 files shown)


data/plugins.json
```diff
@@ -25,7 +25,7 @@
     "chrome": 44,
     "edge": 12,
     "firefox": 34,
-    "safari": 10
+    "safari": 7
```
hzoo: screen shot 2016-10-06 at 2 27 11 pm

data/plugins.json
```diff
@@ -34,12 +34,10 @@
   },
   "transform-es2015-destructuring": {
     "chrome": 51,
-    "edge": 14,
```
hzoo: screen shot 2016-10-06 at 2 30 22 pm

Example of what I mean where it's not 100%

hzoo:
It's only setting node for some of the plugins - I'm finding some missing data on the tables even though the website has the fields as yes/green https://github.com/kangax/compat-table/pull/932

hzoo:
Will merge for now but we need to work on fixing up compat-table data still https://github.com/babel/babel-preset-env/issues/7

PR #19, opened 2016-10-07
Adds `browsers` property to use browserslist's queries

yavorsky:
This PR allows us to use **browsers** property with query like autoprefixer does.

For example, **babelrc**:

```js
{
  presets: [
    ["env", {
      "targets": {
        "chrome": 52,
        "browsers": ["safari > 7"]
      },
      "loose": true
    }]
  ]
}
```
will include chrome 52 and safari 8, 9, 10.

ios_saf, ie11 and others unsupported browsers will be ignored.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,8 @@
     "babel-plugin-transform-es2015-typeof-symbol": "^6.6.0",
     "babel-plugin-transform-es2015-unicode-regex": "^6.3.13",
     "babel-plugin-transform-exponentiation-operator": "^6.8.0",
-    "babel-plugin-transform-regenerator": "^6.6.0"
+    "babel-plugin-transform-regenerator": "^6.6.0",
+    "browserslist": "^1.4.0"
   },
   "devDependencies": {
     "babel-cli": "^6.11.4",
@@ -52,7 +53,6 @@
     "babel-plugin-transform-flow-strip-types": "^6.8.0",
     "babel-preset-es2015": "^6.14.0",
     "babel-register": "^6.14.0",
-    "browserslist": "^1.4.0",
     "compat-table": "github:hzoo/compat-table#node-fix",
     "eslint": "^3.3.1",
     "eslint-config-babel": "^1.0.1",
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "babel-plugin-transform-flow-strip-types": "^6.8.0",
     "babel-preset-es2015": "^6.14.0",
     "babel-register": "^6.14.0",
+    "browserslist": "^1.4.0",
     "compat-table": "github:hzoo/compat-table#node-fix",
     "eslint": "^3.3.1",
     "eslint-config-babel": "^1.0.1",
--- a/src/index.js
+++ b/src/index.js
@@ -40,6 +40,10 @@ export const MODULE_TRANSFORMATIONS = {
  * @return {Boolean}  Whether or not the transformation is required
  */
 export const isPluginRequired = (supportedEnvironments, plugin) => {
+  if (supportedEnvironments.browsers) {
+    supportedEnvironments = getTargets(supportedEnvironments);
+  }
+
   const targetEnvironments = Object.keys(supportedEnvironments);
 
   if (targetEnvironments.length === 0) { return true; }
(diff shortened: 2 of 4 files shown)


hzoo:
awesome!

package.json
```diff
@@ -52,6 +52,7 @@
     "babel-plugin-transform-flow-strip-types": "^6.8.0",
     "babel-preset-es2015": "^6.14.0",
     "babel-register": "^6.14.0",
+    "browserslist": "^1.4.0",
```
hzoo: This will have to be a dependency if we are using it in `src`

src/index.js
```diff
@@ -62,7 +63,16 @@ export const isPluginRequired = (supportedEnvironments, plugin) => {
 };
 
 const getTargets = targetOpts => {
-  return targetOpts || {};
+  const mergedOpts = targetOpts || {};
+  const browserOpts = targetOpts['browsers'];
```
hzoo: Do we want to do any validation here? (like in loose/modules we verify if boolean etc)

hzoo: Oh I see it's ignored right below

package.json
hzoo: Although it would be better if it was used as a devDependency

yavorsky:
I think, browsers defined in **targets**' root must have higher priority than items we receiving in **browsers** query.
So, 
```js
{
  targets: {
    chrome: 50,
    safari: 10,
    browsers: ['last 2 chrome versions', 'last 2 safari versions'] // chrome 52, safari 9
  }
}
```
will output `{chrome: 50, safari: 10}`.

hzoo:
Yep I think so to, we should update the doc/comment to say that explicit targets will override what's in browsers

src/index.js
```diff
@@ -61,8 +66,38 @@ export const isPluginRequired = (supportedEnvironments, plugin) => {
...
-  return targetOpts || {};
+  let mergedOpts;
+  const browserOpts = targetOpts.browsers;
+  if (isBrowsersQueryValid(browserOpts)) {
+    const queryBrowsers = getLowestVersions(browserslist(browserOpts));
+    mergedOpts = mergeBrowsers(queryBrowsers, targetOpts);
+  } else {
+    mergedOpts = targetOpts;
+  }
+  return mergedOpts;
```
hzoo: You could also do an early return here

```
const browserOpts = targetOpts.browsers;
if (isBrowsersQueryValid(browserOpts)) {
  const queryBrowsers = getLowestVersions(browserslist(browserOpts));
  return mergeBrowsers(queryBrowsers, targetOpts);
}
return targetOps;
```

src/index.js
```diff
@@ -39,6 +40,10 @@ export const MODULE_TRANSFORMATIONS = {
  * @return {Boolean}  Whether or not the transformation is required
  */
 export const isPluginRequired = (supportedEnvironments, plugin) => {
+  if (supportedEnvironments.browsers) {
```
yavorsky: @hzoo It's just for usage from a separate module. When options are not handled yet.

hzoo:
Ok great, looks good!! 🎉

chicoxyzzy:
Do we really need possibility to mention browsers directly in targets? AFAIK browserlist can handle this.


```
{
  targets: {
    chrome: 50,
    safari: 10,
    browsers: ['last 2 chrome versions', 'last 2 safari versions'] // chrome 52, safari 9
  }
}
```
is equal to
```
{
  targets: {
    browsers: ['chrome 50', 'safari 10']
  }
}
```

Example: http://browserl.ist/?q=last+2+versions%2C+not+ie+%3C+11%2C+chrome+5
Do we need this only for Node.js?

hzoo:
I guess it depends on what browserlist supports compared to the compat-table - yeah node?

ianwremmel:
It looks like autoprefixer is now [recommending a browserlist file](https://github.com/postcss/autoprefixer#no-prefixes-in-production) in the project root rather than configure tools directly. Does this support that?

hzoo:
Nope it's just a string atm - there https://github.com/babel/babel-preset-env/issues/26 open though!

ianwremmel:
gotcha. thanks @hzoo!

PR #24, opened 2016-10-12
Add Caveat section to Readme.md

basicdays:
The Caveat section details some known issues and workaround when targeting specific environments and using specific plugins.


hzoo:
Thanks added some newlines

PR #30, opened 2016-11-02
If babel doesn't implement it either, don't count against it

hzoo:
Didn't get to this at the time and the fix is actually pretty simple I think?

@jdalton mentioned in https://github.com/babel/babel-preset-env/commit/b80dfaaa7f88cd569b037bd54fc1712329af13d1#commitcomment-19652847 that node 4 supports arrow-functions. It's a similar thing for other transforms which I noticed before but didn't look into. This seems to change a bunch to node 4.

<details>
<summary> the entries babel is failing at (or w/ non-spec option)</summary>
arrow functions/correct precedence

arrow functions/no "prototype" property
arrow functions/lexical "super" binding in constructors
arrow functions/lexical "new.target" binding
class/computed names, temporal dead zone
class/extends null
class/new.target
super/statement in constructors
super/expression in constructors
super/constructor calls use correct "new.target" binding
super/super() invokes the correct constructor
destructuring, parameters/new Function() support
destructuring, parameters/defaults, separate scope
destructuring, parameters/defaults, new Function() support
function "name" property/new Function
function "name" property/bound functions
function "name" property/accessor properties
function "name" property/symbol-keyed methods
function "name" property/object methods (class)
function "name" property/isn't writable, is configurable
Unicode code point escapes/in identifiers
default function parameters/temporal dead zone
default function parameters/separate scope
default function parameters/new Function() support
rest parameters/can't be used in setters
rest parameters/new Function() support
spread (...) operator/with sparse arrays, in array literals
spread (...) operator/spreading non-iterables is a runtime error
RegExp "y" and "u" flags / "y" flag
RegExp "y" and "u" flags / "y" flag, lastIndex
template literals/toString conversion
generators/can't use "this" with new
generators/%GeneratorPrototype%.constructor
generators/yield * on non-iterables is a runtime error
<details>

### DIFF
--- a/data/plugins.json
+++ b/data/plugins.json
@@ -1,10 +1,10 @@
 {
   "transform-es2015-arrow-functions": {
-    "chrome": 47,
+    "chrome": 45,
     "edge": 13,
     "firefox": 45,
     "safari": 10,
-    "node": 6,
+    "node": 4,
     "ios": 10
   },
   "transform-es2015-block-scoped-functions": {
@@ -23,11 +23,11 @@
     "ios": 10
   },
   "transform-es2015-classes": {
-    "chrome": 46,
+    "chrome": 45,
     "edge": 13,
     "firefox": 45,
     "safari": 10,
-    "node": 5,
+    "node": 4,
     "ios": 10
   },
   "transform-es2015-computed-properties": {
@@ -66,21 +66,23 @@
   "transform-es2015-literals": {
     "chrome": 44,
     "edge": 12,
+    "firefox": 40,
     "safari": 9,
     "node": 4,
     "ios": 9
   },
   "transform-es2015-object-super": {
-    "chrome": 46,
+    "chrome": 41,
     "edge": 13,
     "firefox": 45,
-    "safari": 10,
-    "node": 5,
-    "ios": 10
+    "safari": 9,
+    "node": 4,
+    "ios": 9
   },
   "transform-es2015-parameters": {
     "chrome": 49,
     "edge": 14,
+    "firefox": 43,
     "safari": 10,
     "node": 6,
     "ios": 10
@@ -102,16 +104,19 @@
     "ios": 10
   },
   "transform-es2015-sticky-regex": {
-    "chrome": 49,
-    "edge": 13,
-    "firefox": 3,
-    "safari": 10,
-    "node": 6,
-    "ios": 10
+    "chrome": -1,
+    "edge": -1,
+    "firefox": -1,
+    "safari": -1,
+    "node": -1,
+    "ie": -1,
+    "android": -1,
+    "ios": -1,
+    "phantom": -1
   },
   "transform-es2015-template-literals": {
     "chrome": 41,
-    "edge": 13,
+    "edge": 12,
     "firefox": 34,
     "safari": 9,
     "node": 4,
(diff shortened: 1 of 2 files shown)


data/plugins.json
```diff
@@ -88,15 +90,18 @@
     "node": 5
   },
   "transform-es2015-sticky-regex": {
-    "chrome": 49,
-    "edge": 13,
-    "firefox": 3,
-    "safari": 10,
-    "node": 6
+    "chrome": -1,
```
hzoo: This one is interesting since it seems as if the transform doesn't actually work?

data/plugins.json
```diff
@@ -1,10 +1,10 @@
 {
   "transform-es2015-arrow-functions": {
-    "chrome": 47,
+    "chrome": 45,
     "edge": 13,
     "firefox": 45,
     "safari": 10,
-    "node": 6
+    "node": 4
```
Kovensky: The `Function.name` will be incorrect, and the `function-name` transform can only fix it for transformed arrow functions.

data/plugins.json
```diff
@@ -20,11 +20,11 @@
     "node": 6
   },
   "transform-es2015-classes": {
-    "chrome": 46,
+    "chrome": 45,
     "edge": 13,
     "firefox": 45,
     "safari": 10,
-    "node": 5
+    "node": 4
```
Kovensky: Node 4 doesn't support spread, so non-transformed classes that have `super(...args)` will not run.

```js
class X extends Y {
  constructor (...args) {
    super(...args)
  }
}
```

is invalid in node 4 regardless of the other transforms. That example constructor is useless as that is the default implementation, but when `transform-class-properties` runs, it has to generate a constructor with that content to attach the public property initializers:

```js
class X extends Y {
  z = null
}
```

transforms to:

```js
class X extends Y {
  constructor (...args) {
    super(...args)
    this.z = null
  }
}
```

data/plugins.json
Kovensky: Though I think that is the case with Node 6 too (but should be fixed in Node 7).

jdalton:
```js
var a = () =>1
a.name // is '' in v4, 'a' in v6 & 7.
```

jdalton: In my arrow function use at least, name wasn't a stopper.  \@sindresorhus

Kovensky: Yeah, it's definitely not a stopper, though it might make things more confusing when debugging, especially for functional React components.

data/plugins.json
hzoo: Is it possible to change spread to account for this? Or do we need to check if a user is using spread..

hzoo: Ok I guess we'll need to figure out how to override this or just not do this PR

Kovensky: I wrote, a long time ago in phabricator (no idea on the bug number), that transform-rest-spread should, at least, detect if super with spread is being used, and bail if transform-es2015-classes is not also being used.

hzoo:
Closing for now, doing this may cause more issues

jdalton:
Is there a way to opt-out of transforms. For example say I don't want to transform arrow functions in Node 4 (I don't care about whatever edge case issue there is).

hzoo:
there's the exclude option

jdalton:
Something like
```json
  "presets": [
    ["env", {
      "exclude": ["transform-es2015-arrow-functions"],
      "targets": { "node": 4 }
    }]
  ]
```

Issue #9, opened 2016-09-24
Auto detect for node


hzoo:
https://github.com/babel/babel/pull/3476#issuecomment-213222405

> Would be super nice if it could just adhere to what's in the [`engines` field](https://github.com/sindresorhus/ava/blob/3a4fc998cb6d361cbc49d3864937c9b18c279764/package.json#L37) in package.json. That way you wouldn't have to specify anything manually.

Or just determine current node version before running?

deepsweet:
> Or just determine current node version before running?

https://github.com/rtsao/babel-preset-es2015-node/blob/master/index.js

hzoo:
yeah we can use `process.version` but we still need to specify whether you are compiling node or compiling for other environments

rtsao:
What about a special value for node (e.g. `{node: "process"}`) which infers from process.version?

AlicanC:
@rtsao great idea. We should support keywords in versions. #23 needs that too.

shubheksha:
I'll work on this!

hzoo:
Awesome - I'l try to merge something of https://github.com/babel/babel-preset-env/pull/19 first to fix merge conflicts

hzoo:
@shubheksha still able to work on this? no rush though

hzoo:
You can do what was in https://github.com/facebookincubator/create-react-app/pull/878 if you use a preset or use a local js file. If you use a package.json/babelrc you won't be able to use javascript in the config so we could have a hardcoded option for this.

so you can specify `node: "process"` which will be the same as `node: parseFloat(process.versions.node)`

And maybe `process` isn't the best option name for this?

hzoo:
@shubheksha Let me know if you need any more context/help on this, or if someone else can take this too

hzoo:
https://github.com/babel/babel-preset-env/issues/35

w/ `"node": "current"` or `"node": true`

PR #35, opened 2016-11-02
hardcode a current node version option

hzoo:
For https://github.com/babel/babel-preset-env/issues/9

If a user wants to compile babel against the current node version (`node --version`) but doesn't want to use js for the config (in package.json/.babelrc rather than say babel-loader or a separate file)

`node: "--version"` -> `"node": parseFloat(process.versions.node)`?

screen shot 2016-11-02 at 2 43 36 pm

Option name ideas?

- true
- "version"
- "current version"
- "process"
- "--version"

cc @shubheksha

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Currently: "chrome, edge, firefox, safari, ie, node".
 
 > Note, browsers' results are overridden by explicit items from `targets`.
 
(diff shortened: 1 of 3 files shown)


mathiasbynens:
`"current"`

Issue #39, opened 2016-11-06
Run tests on OSX on Travis


sheerun:
My experience suggests that babel-preset-env has issues with async/awain on node 4, only on osx. It is shown here: https://travis-ci.org/sheerun/modern-node/builds/173713418

Could you please enable OSX builds on Travis, so you can avoid such issues in the future.

I also hope this issue can be fixed as well, after Travis is setup.

hzoo:
Do you want to make a PR to add osx? I haven't done it before

sheerun:
Just add following to .travis.yml:

```
os:
  - osx
  - linux
```

PR #40, opened 2016-11-08
Add OS X builds to Travis

hawkrives:
This first attempt might fail, but I've gotten it working before.

(For #39)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,19 +5,19 @@ language: node_js
 cache:
   directories:
   - node_modules
+node_js:
+- '7'
+- '6'
+- '4'
+- '0.12'
+- '0.10'
 script:
   - 'if [ -n "${LINT-}" ]; then npm run lint ; fi'
   - 'if [ -z "${LINT-}" ]; then npm run ci ; fi'
-node_js:
-  - '7'
-  - '6'
-  - '4'
-  - '0.12'
-  - '0.10'
 os: [linux, osx]
 matrix:
   fast_finish: true
   include:
-    - os: linux
-      node_js: "node"
+    - node_js: "node"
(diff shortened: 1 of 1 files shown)


hawkrives:
The downside to testing on OS X is that Travis's OS X build queue is usually full:

screen shot 2016-11-08 at 11 33 58 am

(It usually fills right back up at around 7am. Today seems to be slow.)

danez:
Maybe we could only run latest node on mac os?

hawkrives:
I thought about that, but #39 mentioned a bug with node@4 on OS X so…

hzoo:
can do allow_failures toohttps://github.com/sheerun/modern-node/blob/master/.travis.yml#L18-L19 cc @sheerun

hawkrives:
If we do `allow_failures`, will the build status here note that? Will anyone ever see when the OS X build fails?

sheerun:
I think you shound't ignore failed osx builds

hzoo:
Although babel-preset-env's own tests are regarding manipulating the plugins and don't actually run anything. We don't have any integration type tests

sheerun:
Well, you could have :)

hzoo:
Most babel tests themselves don't actual do an exec test - mostly a string check that `transform(actual) === expected`

sheerun:
Ideally there would be an example app that uses all of es6 features, you compile it for different environments, and run it there. Maybe @browserstack or @saucelabs could sponsor

hzoo:
Were using the data from compat-table so we trust that it is correct on there but yeah would be nice!

hzoo:
Let's come back to this when we setup exec tests if we need to

PR #43, opened 2016-11-10
Rename ios_saf in browserslist data to ios, ignore unknown browsers

Kovensky:
browserslist queries, especially ones with percentages, often bring in mobile browsers that _probably_ have the similar support level as desktop ones, but since there's no support for them in the data, including them here would just lead to all plugins being enabled always.

It's also easy to get crazy things like `operamini` or `and_uc` in the results even if you don't actually want to support them.

Browserslist also reports iOS Safari as `ios_saf`, while the data uses `ios`, so it needs to be renamed.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -48,13 +48,13 @@ const isBrowsersQueryValid = (browsers) => {
 };
 
 const browserNameMap = {
-  chrome: 'chrome',
-  edge: 'edge',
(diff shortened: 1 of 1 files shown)


hzoo:
Maybe we should be mentioning this in the docs so folks know

hzoo:
👍

PR #44, opened 2016-11-14
Update data generation for latest compat-table

chicoxyzzy:
Now all environments data is located in `environments.js`. It is still available under `es6data.browsers` and `esPlusData.browsers` but `browsers` properties will be removed from test results in future.

See https://github.com/kangax/compat-table/pull/960


hzoo:
great 👍

PR #46, opened 2016-11-16
Only console.log the debug info once

hzoo:
Saw that when I was running the debug option it would be logging many times which isn't necessary.


noinkling:
Would it be possible to scope this flag somehow for the situation described in #147 ?

PR #50, opened 2016-11-20
Update data generation for latest compat-table

chicoxyzzy:
See https://github.com/kangax/compat-table/pull/964

Still need to realise why ios 8 was removed from new generated data

### DIFF
--- a/data/plugins.json
+++ b/data/plugins.json
@@ -19,6 +19,7 @@
   },
   "transform-es2015-block-scoping": {
     "chrome": 49,
+    "opera": 36,
     "firefox": 51,
     "safari": 10,
     "node": 6,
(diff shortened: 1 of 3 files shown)


chicoxyzzy:
@hzoo just checked latest `master` branch and I'm getting `ios` removed for some plugins on it. Same as in my branch.

I've tried:
- node v4.4.4 (npm v3.9.0)
- node v6.9.1 (npm v4.0.0)
- node v7.1.0 (npm v4.0.2)
- node v7.2.0 (npm v4.0.3)

Results are always the same — `ios` removed for same plugins on both branches

chicoxyzzy:
OK. Seems like it caused by recent changes in `compat-table`'s master and you've built https://github.com/babel/babel-preset-env/pull/48 using not the latest one

chicoxyzzy:
compat-table version is fixed to particular commit now so this PR should fix `build-data` task for now.

Currently I'm working on making raw data available (follow #965) so we'll be able to use it instead of generating it in babel-preset-env

PR #53, opened 2016-11-28
add support for engines with flags Fixes #41

### LINKED ISSUE #41: support node71-harmony
Allow specifying node versions with harmony flags.

graingert:
If you're targeting node --harmony you can use node-flags: 6.5

### DIFF
--- a/data/plugins.json
+++ b/data/plugins.json
@@ -6,6 +6,9 @@
     "safari": 10,
     "node": 6,
     "ios": 10,
+    "chrome-flag": 47,
+    "edge-flag": 13,
+    "node-flag": 6,
     "opera": 34
   },
   "transform-es2015-block-scoped-functions": {
@@ -15,6 +18,8 @@
     "node": 4,
     "ie": 11,
     "ios": 10,
+    "chrome-flag": null,
+    "node-flag": 0.12,
     "opera": 28
   },
   "transform-es2015-block-scoping": {
@@ -23,6 +28,8 @@
     "safari": 10,
     "node": 6,
     "ios": 10,
+    "chrome-flag": 48,
+    "node-flag": 6,
     "opera": 36
   },
   "transform-es2015-classes": {
@@ -32,6 +39,9 @@
     "safari": 10,
     "node": 5,
     "ios": 10,
+    "chrome-flag": 46,
+    "edge-flag": 13,
+    "node-flag": 5,
     "opera": 33
   },
   "transform-es2015-computed-properties": {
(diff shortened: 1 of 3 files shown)


hzoo:
Thanks.

Not sure we want to do it like this with x-flag? We need to update the readme too

graingert:
@hzoo I didn't update the docs to keep it as a hidden feature for now.

data/plugins.json
```diff
@@ -157,23 +206,33 @@
...
     "opera": 39
   },
   "transform-async-to-generator": {
     "chrome": 55,
     "firefox": 52,
+    "chrome-flag": 52,
+    "edge-flag": 13,
     "opera": 42
   },
   "syntax-trailing-function-commas": {
```
Kovensky: (not related to this specific PR but) wait... a syntax plugin? Isn't this only used for enabling `babylon` features?

Or does it happen to work because of how it interacts with `babel-generate`? If that's the case, should not preserving the trailing comma be considered a `babel-generate` bug? 😄

data/plugins.json
```diff
@@ -157,23 +206,33 @@
     "safari": 10,
     "node": 6,
     "ios": 10,
+    "chrome-flag": null,
```
Kovensky: This needs to be (at least) `56` because of a bug where `return` in generators could be converted into tail calls when `--harmony` is enabled, causing nonsense crashes.

data/plugins.json
```diff
@@ -157,23 +206,33 @@
...
+    "edge-flag": 13,
+    "node-flag": 6,
```
Kovensky: There's no node release with v8 5.6, which fixes the aforementioned tail call crash.

data/plugins.json
hzoo: This is allowing the user to add the trailing comma (babel-generator always removes trailing commas). Without the plugin it will error on unexpected token?

However, I just realized after your comment that we removed this restriction in babylon since its part of es2017 so I guess we don't really need this plugin anymore after a certain babylon version unless generator prints trailing commas by default later (no)

data/plugins.json
graingert: that's definitely a problem with compat-table's data then

data/plugins.json
graingert: Ideally I'd like to keep the trailing comma, both for technical purity reasons and for making the generated code as close as possible to what's written

hzoo: Yeah I would too

graingert:
@hzoo hmm I thought this was passing before...

graingert:
@hzoo there we go. Build working now

hzoo:
Thanks for the PR! 

PR #55, opened 2016-12-02
Support Electron in babel-preset-env

paulcbetts:
Map Electron versions onto Chromium versions

## TODO:

- [ ] Get tests running
- [ ] Code Review


hzoo:
So fast 👌

src/index.js
```diff
@@ -79,11 +80,31 @@ export const getCurrentNodeVersion = () => {
...
+
 export const getTargets = (targetOpts = {}) => {
   if (targetOpts.node === true || targetOpts.node === "current") {
     targetOpts.node = getCurrentNodeVersion();
   }
 
+  // Rewrite Electron versions to their Chrome equivalents
+  if (targetOpts.electron) {
+    targetOpts.chrome = electronVersionToChromeVersion(targetOpts.electron);
+    delete targetOpts.electron;
```
hzoo: Do you think it's better to delete the option? I guess it makes sense still given it's using the chrome version anyway

data/electronToChromium.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  "1.5": 54,
```
hzoo: Is this the best way to track versions or is there a place for this already? cc @kevinsawicki

kevinsawicki: I think it is okay to track this here. there is https://atom.io/download/electron/index.json which contains all the specific versions so this file could be generated from that.

kevinsawicki: Since 1.5 is not out yet, and depending on how long it takes to upgrade Electron to Chrome 54, Electron may skip it 54 altogether and go right to Chrome 55 which is due to hit stable very soon.

hzoo: All good - we can update and make a new version whenever, thanks for the head's up.

hzoo: @kevinsawicki let me know how we can help to get atom to use this (if you're interested)

hzoo:
Alright fixed the bug + added tests. Thanks @paulcbetts!!

amilajack:
What exactly does `useBuiltins` to? I'm still confused about this

hzoo:
@amilajack this pr is really related to useBuiltIns?

If you just have an usage issue or want help, I would recommend checking out our [slack community](https://slack.babeljs.io/) or [StackOverflow](http://stackoverflow.com/questions/tagged/babeljs).

https://github.com/babel/babel-preset-env#usebuiltins it transforms babel-polyfill to the individual polyfills per env

paulcbetts:
@hzoo going to add this into electron-forge (https://github.com/electron-userland/electron-forge/pull/47), so babel-preset-env will hopefully be in-box when people create new projects. Thanks again for the great plugin!

Issue #20, opened 2016-10-08
Feature request: transform a require('babel-polyfill') call to specific requires base on the env config


keyanzhang:
First of all, `babel-preset-env` is awesome. Thanks for making this happen!

I'm thinking about how to make polyfills better. If someone uses `babel-preset-env` it means he/she has a few specific platforms to target and it's probably not necessary to require the entire `babel-polyfill` library. 

Since we now know what the targets are, can we transform a `require('babel-polyfill');` call to specific requires? For example, if someone targets IE 11 then 

``` js
require('babel-polyfill');
```

gets transformed to

``` js
require('babel-polyfill/array-from');
require('babel-polyfill/array-of');
require('babel-polyfill/object-assign');
require('babel-polyfill/regenerator-runtime');
/* ... other polyfills that IE 11 needs */
```

The example above is not comprehensive and there should be a better way to require individual polyfill libraries, but you get the idea. If someone targets Chrome 55 then there are only very few polyfills that need to be applied.

It's probably gonna be a new transform plugin (`transform-polyfill-require`? :D) and won't live in this repo, but it should use the compat-table data here. Feedbacks are welcome!

Related tweet: https://twitter.com/left_pad/status/784611480947810304

hzoo:
Yeah I guess this would be a very specific plugin since I would think you only have `require('babel-polyfill');/import` once in your app (at the entry point). Crazy but we should try it out to see!

ljharb:
The reality is that every browser needs every polyfill, because they all have quirks in their implementations that need correcting.

tizmagik:
> The reality is that every browser needs every polyfill, because they all have quirks in their implementations that need correcting.

I'm not sure that's true, but regardless, a polyfill wouldn't solve that anyway since most polyfills are defined like so (e.g. only polyfill if there is no native implementation):

```js
Array.prototype.includes = Array.prototype.includes || function includesPolyfill() { ... }
```

ljharb:
Imperfect ones, sure. Polyfills/shims of sufficient quality actually runtime-test their implementation (when a buggy one exists) to ensure it behaves properly: https://github.com/ljharb/object.assign/blob/master/polyfill.js#L41-L50

tizmagik:
Gotcha, thanks for the clarification @ljharb -- it looks like corejs (which babel-polyfill uses) does do some runtime implementation checking in some cases, so that's good to know! 👍

Munter:
[...]

PR #56, opened 2016-12-02
add useBuiltIns option

hzoo:
Ref https://github.com/babel/babel-preset-env/issues/20

TD;LR:

turns 

```js
import 'babel-polyfill'; // all of the polyfill
```

into 

```js
import "core-js/modules/es6.object.assign"; // required individual modules
import "core-js/modules/...";`
...
```

based on supported environment

screen shot 2016-12-01 at 8 34 51 pm

cc @zloirock, @chicoxyzzy, @Kovensky, @keyanzhang, @ljharb

compat-table data -> data/builtInFeatures.js (mapping core-js module to builtins) -> babel plugin that turns `import babel-polyfill` into individual imports

---
- [x] add the rest of the built-ins (there's a lot)
- [x] verify that the mapping for polyfill-> compat-table is correct
  - example: "typed/array-buffer"/typed arrays - should it be per array or just the whole feature?
  - references are https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-runtime/src/definitions.js
  - http://kangax.github.io/compat-table/es6
  - https://github.com/zloirock/core-js
- [x] create the option in the preset to turn this on
- [x] write the babel plugin to transform the import/require statement for `import babel-polyfill`
- [x] test the plugin with http://henryzoo.com/esprima_ast_explorer/#/ZOw13NK8Tr/4
- [x] readme update
- [x] tests
- [x] fix empty data bug
- [x] prevent duplicate imports
- [ ] what to do about experimental built-ins and compatibility with the regular `import "babel-polyfill"`?
- [x] regenerator
---

` data/builtInFeatures.js` is similar to `data/pluginFeatures.js`

We try to match the [core-js](https://github.com/zloirock/core-js#commonjs) module name with the corresponding test name in the [compat-table](https://github.com/kangax/compat-table/blob/gh-pages/data-es6.js).

```
"es6.typed.int8-array": "typed arrays / Int8Array",
```

Example:

screen shot 2016-12-02 at 5 09 10 pm

```
"es6.object.assign": "Object static methods / Object.assign",
"es6.object.is": "Object static methods / Object.is",
```

`import 'babel-polyfill'` -> `import "core-js/modules/es6.object.assign";`


hzoo:
via @zloirock

`symbol/to-string-tag` -> `es6.object.to-string`
`symbol/match` and related -> `es6.regexp.match` etc
`es6.array.iterator` for array.prototype[values|keys|entries]

data/builtInFeatures.js
```diff
@@ -0,0 +1,120 @@
+// https://github.com/zloirock/core-js
+
+/* eslint-disable quotes */
```
Kovensky: You can just set it with `avoidEscape`

http://eslint.org/docs/rules/quotes#options

scripts/build-data.js
```diff
@@ -115,27 +116,47 @@ const getLowestImplementedVersion = ({ features }, env) => {
...
-    // add opera
-    if (plugin.chrome) {
-      plugin.opera = plugin.chrome - 13;
-    }
+      // add opera
+      if (plugin.chrome) {
+        plugin.opera = plugin.chrome - 13;
+
+        if (plugin.chrome === 5) {
+          plugin.opera = 12;
```
Kovensky: This check is confusing, and probably doesn't do what it intended to.

Maybe `if (plugin.chrome && plugin.chrome >= 28)`, with the special case for chrome 5 later?

hzoo: I think it's just b/c I found opera 12/chrome 5 hardcoded but there weren't any opera 13,14 etc

Kovensky: Because of how the statement in line 140 sets opera to chrome - 13, you might get opera -12, opera -11, ..., opera 14 (assuming there is data for chrome older than 28) :X

data/builtInFeatures.js
```diff
@@ -0,0 +1,120 @@
...
+module.exports = {
+  // es2015
+  // core-js/fn/map
+
+  // "es6.typed/array-buffer": "typed arrays / ",
+  "es6.typed.data-view": "typed arrays / DataView",
```
hzoo: this should cover

typed arrays / DataView (Int8)
typed arrays / DataView (Uint8)
typed arrays / DataView (Int16)
typed arrays / DataView (Uint16)
typed arrays / DataView (Int32)
typed arrays / DataView (Uint32)
typed arrays / DataView (Float32)
typed arrays / DataView (Float64)

data/builtInFeatures.js
```diff
@@ -0,0 +1,120 @@
+// https://github.com/zloirock/core-js
+
+/* eslint-disable quotes */
+module.exports = {
+  // es2015
+  // core-js/fn/map
+
+  // "es6.typed/array-buffer": "typed arrays / ",
```
hzoo: I think these would be covered with all the `int8-array` ones below

typed arrays / ArrayBuffer[Symbol.species]
typed arrays / constructors require new
typed arrays / constructors accept generic iterables
typed arrays / correct prototype chains
typed arrays / %TypedArray%.from
typed arrays / %TypedArray%.of
typed arrays / %TypedArray%.prototype.subarray
typed arrays / %TypedArray%.prototype.join
typed arrays / %TypedArray%.prototype.indexOf
typed arrays / %TypedArray%.prototype.lastIndexOf
typed arrays / %TypedArray%.prototype.slice
typed arrays / %TypedArray%.prototype.every
typed arrays / %TypedArray%.prototype.filter
typed arrays / %TypedArray%.prototype.forEach
typed arrays / %TypedArray%.prototype.map
typed arrays / %TypedArray%.prototype.reduce
typed arrays / %TypedArray%.prototype.reduceRight
typed arrays / %TypedArray%.prototype.reverse
typed arrays / %TypedArray%.prototype.some
typed arrays / %TypedArray%.prototype.sort
typed arrays / %TypedArray%.prototype.copyWithin
typed arrays / %TypedArray%.prototype.find
typed arrays / %TypedArray%.prototype.findIndex
typed arrays / %TypedArray%.prototype.fill
typed arrays / %TypedArray%.prototype.keys
typed arrays / %TypedArray%.prototype.values
typed arrays / %TypedArray%.prototype.entries
typed arrays / %TypedArray%.prototype[Symbol.iterator]
typed arrays / %TypedArray%[Symbol.species]

data/builtInFeatures.js
```diff
@@ -0,0 +1,119 @@
...
+
+  "es6.array.from": "Array static methods / Array.from",
+  "es6.array.of": "Array static methods / Array.of",
+
+  "es6.array.copy-within": "Array.prototype methods / Array.prototype.copyWithin",
+  "es6.array.find": "Array.prototype methods / Array.prototype.find",
+  "es6.array.find-index": "Array.prototype methods / Array.prototype.findIndex",
+  "es6.array.fill": "Array.prototype methods / Array.prototype.fill",
+  "es6.array.iterator": "Array.prototype methods / Array.prototype.keys",
+  // "es6.array.iterator": "Array.prototype methods / Array.prototype.values",
```
weikinhuang: perhaps a note about these

README.md
```diff
@@ -120,6 +120,39 @@ Useful if there is a bug in a native implementation, or a combination of a non-s
...
+
+A way to apply `babel-preset-env` for polyfills (via "babel-polyfill").
+
+```
+npm install babel-polyfill --save
+```
+
+This option will apply a new plugin that replaces the statement `import "babel-polyfill"` or `require("babel-polyfill")` with individual requires for `babel-polyfill` based on environment.
+
+> NOTE: This means you should only use `require("babel-polyfill");` once in your whole app. I would recommend just creating a single file that only does the require.
```
kaicataldo: Nit: I generally prefer not to use "I" in readmes unless there's a reason you want it to be explicit that you, the author, are saying something. Maybe something like, "Creating a single file that only does contains the require is recommended."?

test/fixtures/preset-options/use-builtins-require/expected.js
```diff
@@ -0,0 +1,13 @@
+require("core-js/modules/es6.typed.data-view");
+
```
kaicataldo: Curious why these have newlines between them

hzoo: me too lol.. probably a generator newline issue

README.md
hzoo: 👍  ok or "One option is to "

src/transformPolyfillRequirePlugin.js
```diff
@@ -0,0 +1,75 @@
+const polyfillSource = "babel-polyfill";
```
hzoo: Oh I guess it should probably work on `import "core-js"` too..

and right now it's weird that it would be change  `import "babel-polyfill"` to "core-js" (since babel-polyfill doesn't have any generated modules atm anyway..)

data/builtInFeatures.js
```diff
@@ -0,0 +1,114 @@
...
+  "es6.reflect.is-extensible": "Reflect / Reflect.isExtensible",
+  "es6.reflect.own-keys": "Reflect / Reflect.ownKeys",
+  "es6.reflect.prevent-extensions": "Reflect / Reflect.preventExtensions",
+  "es6.reflect.set": "Reflect / Reflect.set",
+  "es6.reflect.set-prototype-of": "Reflect / Reflect.setPrototypeOf",
+
+  "es6.promise": "Promise",
+  "es6.symbol": "Symbol",
+
+  "es6.symbol": "well-known symbols / Symbol.hasInstance",
```
hzoo: `@@hasInstance` - works with `babel` optional transform + `core-js`
`@@isConcatSpreadable` - can be polyfilled, but for some performance reason I didn't add it before. Maybe if I'll find some time in the future...
`@@iterator` - works with `core-js` and some `babel` transforms.
`@@match` - works with `core-js`
`@@replace` - works with `core-js`
`@@search` - works with `core-js`
`@@species` - in some places works with `core-js`
`@@split` - works with `core-js`
`@@toPrimitive` - can't be completely polyfilled / transpiled
`@@toStringTag` - works with `core-js`
`@@unscopables` - theoretically, can be transpiled, but it should work in sloppy mode only

from zloirock on slack

data/builtIns.json
```diff
@@ -0,0 +1,726 @@
...
+  },
+  "es6.reflect": {
+    "chrome": 49,
+    "opera": 36,
+    "edge": 13,
+    "firefox": 45,
+    "safari": 10,
+    "ios": 10
+  },
+  "es6.reflect.apply": {},
```
Kovensky: Missing data

data/builtIns.json
```diff
@@ -0,0 +1,726 @@
...
+  "es6.object.get-own-property-symbols": {
+    "chrome": 38,
+    "opera": 25,
+    "edge": 12,
+    "firefox": 36,
+    "safari": 9,
+    "node": 0.12,
+    "ios": 9
+  },
+  "es6.object.set-prototype-of": {
```
Kovensky: Oh, right, `__proto__` used to be a gecko extension

hzoo:
@Kovensky https://github.com/babel/babel-preset-env/pull/62 fixes missing data

hzoo:
https://unpkg.com/babel-polyfill@6.16.0/lib/index.js

Now i'm also wondering about the compatibility and differences with the regular import of babel-polyfill.

I see in `require("core-js/shim");` it requires experimental features as well but this option will not so there's a difference there.

hzoo:
also need to deal with regeneratorRuntime

Kovensky:
`core-js/shim` also includes `web.timers` (only needed for IE <= 9?), `web.immediate` (experimental?) and `web.dom.iterable`.

hzoo:
Should we just auto include all the experimental ones?

Kovensky:
At least `web.dom.iterable` is certainly not experimental, and definitely should be included for DOM/CSSOM compliance, but it also doesn't need to be included in node-only builds (unless you expect to patch jsdom?).

The problem with `web.dom.iterable`, though, is that I don't know what data can be used to tell which browsers support it 😕 

I'm also not quite sure on including experimentals. On one hand, `env` tries to provide `latest`-level syntax, but should it really be providing things that go beyonde `latest` if they are API? If they are experimental, how do you know when/where to omit them?

On the other hand, you might break people that were using the experimental APIs without noticing. Adding an `experimental` option could work, but that could be interpreted to mean stage-(n < 4) features...

hzoo:
Yeah I would like to do add any experimentals. So you think `web.dom.iterable` is the only one we should also add @Kovensky?

data/builtInFeatures.js
```diff
@@ -0,0 +1,125 @@
...
+  "es6.symbol": "well-known symbols / Symbol.isConcatSpreadable",
+  "es6.symbol": "well-known symbols / Symbol.iterator",
+  "es6.symbol": "well-known symbols / Symbol.match",
+  "es6.symbol": "well-known symbols / Symbol.replace",
+  "es6.symbol": "well-known symbols / Symbol.search",
+  "es6.symbol": "well-known symbols / Symbol.species",
+  "es6.symbol": "well-known symbols / Symbol.split",
+  "es6.symbol": "well-known symbols / Symbol.toPrimitive",
+  "es6.symbol": "well-known symbols / Symbol.toStringTag",
+  "es6.symbol": "well-known symbols / Symbol.unscopables",
```
zloirock: Duplicated object keys is not a problem?

hzoo: Hmm good point it is..

hzoo: Maybe just should remove that section

zloirock: As I wrote, for me, the structure like `"core-js.module.name": ["test 1", "test 2", ...]` looks like the best option here.

data/builtInFeatures.js
```diff
@@ -0,0 +1,125 @@
...
+  "es6.array.of": "Array static methods / Array.of",
+
+  "es6.array.copy-within": "Array.prototype methods / Array.prototype.copyWithin",
+  "es6.array.find": "Array.prototype methods / Array.prototype.find",
+  "es6.array.find-index": "Array.prototype methods / Array.prototype.findIndex",
+  "es6.array.fill": "Array.prototype methods / Array.prototype.fill",
+  "es6.array.iterator": "Array.prototype methods / Array.prototype.keys",
+  // can use Symbol.iterator, not implemented in many environments
+  // "es6.array.iterator": "Array.prototype methods / Array.prototype.values",
+  "es6.array.iterator": "Array.prototype methods / Array.prototype.entries",
```
existentialism: Just a note, another case of dupe keys similar to `es6.symbol`

data/builtInFeatures.js
yavorsky: Or maybe to prevent hardcoding just use it with RegExp:
For `s6.symbol`: `/(^|\s)Symbol(\.\w+)?$/` will match `Symbol` and `/ Symbol.method`.
Or `/Array.prototype.(keys|entries)$/` for `es6.array.iterator`.
But It might cause conflicts in the future. Needs really accurate RegExps.

zloirock: Many modules should have in requirements some tests from different categories, so just a list of tests should be better.

hzoo: Alright yeah I thought it was be "cool" to do it this way but having it be the same as pluginFeatures is good

data/builtInFeatures.js
```diff
@@ -0,0 +1,134 @@
...
+  "es6.typed.int32-array": "typed arrays / Int32Array",
+  "es6.typed.uint32-array": "typed arrays / Uint32Array",
+  "es6.typed.float32-array": "typed arrays / Float32Array",
+  "es6.typed.float64-array": "typed arrays / Float64Array",
+
+  "es6.map": "Map",
+  "es6.set": "Set",
+  "es6.weak-map": "WeakMap",
+  "es6.weak-set": "WeakSet",
+  // proxy not implemented
```
hzoo: right bad wording haha

data/builtInFeatures.js
```diff
@@ -0,0 +1,134 @@
...
+  "es6.string.from-code-point": "String static methods / String.fromCodePoint",
+
+  "es6.string.code-point-at": "String.prototype methods / String.prototype.codePointAt",
+  // "String.prototype methods / String.prototype.normalize" not implemented
+  "es6.string.repeat": "String.prototype methods / String.prototype.repeat",
+  "es6.string.starts-with": "String.prototype methods / String.prototype.startsWith",
+  "es6.string.ends-with": "String.prototype methods / String.prototype.endsWith",
+  "es6.string.includes": "String.prototype methods / String.prototype.includes",
+
+  "es6.regexp.flags": "RegExp.prototype properties / RegExp.prototype.flags",
```
hzoo: You mean for http://babeljs.io/docs/plugins/transform-es2015-unicode-regex/#installation?

(we need an example there)

src/index.js
```diff
@@ -134,38 +136,58 @@ export const validateWhitelistOption = (whitelistOpt = []) => {
...
+  console.log(logStr);
+};
+
 export default function buildPreset(context, opts = {}) {
   const loose = validateLooseOption(opts.loose);
   const moduleType = validateModulesOption(opts.modules);
   const whitelist = validateWhitelistOption(opts.whitelist);
   const targets = getTargets(opts.targets);
   const debug = opts.debug;
+  const useBuiltIns = opts.useBuiltIns;
```
hzoo: @kaicataldo asked about this as well.

It is an implementation + it's just the name we used before that's all.

https://github.com/babel/babel/pull/3327 and we went with `useBuiltIns`

scripts/build-data.js
```diff
@@ -98,7 +114,7 @@ const getLowestImplementedVersion = ({ features }, env) => {
   });
 
   let envFiltered = envTests.filter((t) => t);
-  if (envTests.length > envFiltered.length) {
+  if (envTests.length > envFiltered.length || envTests.length === 0) {
```
hzoo: will just leave it for now

hzoo:
Alright lets give it a shot and move forward - will fix if we find any issues.

PR #59, opened 2016-12-05
Add logs for polyfills. 

yavorsky:
1. Added `Using polyfills:` to log when debug is enabled.
2. Log module before `Using plugins:` row.


hzoo:
Not sure about committing the lock file for the project but that's just me? Is that what we're all doing now? Seems weird because a consumer/user could have a later version with `^` that is broken and we wouldn't know unless we have greenkeeper + yarn

hzoo:
I see the CI auto picks it up with the lock file but yarn only works on node 4 and 0.10/0.12 still uses npm so that concern is not a big deal unless we drop those.

yavorsky:
@hzoo yes, nice point! Moreover it must be added with the separate PR. Just forgot about it.

hzoo:
Given we are doing node 0.10 still we are covered so yeah another pr

src/index.js
```diff
@@ -136,6 +136,17 @@ export const validateWhitelistOption = (whitelistOpt = []) => {
...
 let hasBeenLogged = false;
 
+const logPlugin = (plugin, targets, list) => {
+  const envList = list[plugin];
+  const filteredList = Object.keys(targets)
+  .reduce((a, b) => {
+    a[b] = envList[b];
+    return a;
+  }, {});
+  const logStr = `\n ${plugin} ${JSON.stringify(filteredList)}`;
```
hzoo: do we need the `\n` given its separate console.logs? I see that it's newlines? (Unless you think it looks better that way)


```
Using plugins:

 syntax-trailing-function-commas {}

Using polyfills:

 es6.typed.data-view {}

 es6.reflect.apply {}

 es6.reflect.own-keys {}

 es6.symbol.iterator {}

 es6.symbol.species {}

 es6.array.from {}
```

```
Using plugins:
syntax-trailing-function-commas {}

Using polyfills:
es6.typed.data-view {}
es6.reflect.apply {}
es6.reflect.own-keys {}
es6.symbol.iterator {}
es6.symbol.species {}
es6.array.from {}
```

hzoo:
Thanks again @yavorsky (added you as a collab to this repo if you'd like)

yavorsky:
@hzoo Thanks! Glad to help you in future!  👍

PR #61, opened 2016-12-06
Fill es6.reflect.own-keys field.

yavorsky:
Last commit just fills the `es6.reflect.own-keys` field, but other changes in *builtIns.json* seems to be https://github.com/babel/babel-preset-env/commit/cfa0f6f1f7612cf2fd01c4af2798ab9906243fb8 and https://github.com/babel/babel-preset-env/commit/c95759d118fa2fc745e7307f3a454c09b19c00c6 reflection.


data/builtInFeatures.js
```diff
@@ -28,7 +28,8 @@ const es2015 = {
   "es6.reflect.get-prototype-of": "Reflect / Reflect.getPrototypeOf",
   "es6.reflect.has": "Reflect / Reflect.has",
   "es6.reflect.is-extensible": "Reflect / Reflect.isExtensible",
-  "es6.reflect.own-keys": "Reflect / Reflect.ownKeys",
```
hzoo: So in pluginFeatures we would have something like https://github.com/babel/babel-preset-env/blob/master/data/pluginFeatures.js#L97-L99
```
  "transform-es2015-unicode-regex": {
    features: [
      'RegExp "y" and "u" flags / "u" flag, case folding',
      'RegExp "y" and "u" flags / "u" flag, Unicode code point escapes',
      'RegExp "y" and "u" flags / "u" flag',
    ],
  },
```

so I changed the format a bit but the logic doesn't work the same

```
const getLowestImplementedVersion = ({ features }, env) => {
  let tests = flatten(compatibilityTests
    .filter((test) => features.indexOf(test.name) >= 0)
```

same with DataView

so features would be ["typed arrays / DataView"]

["typed arrays / DataView"].indexOf(test.name) >= 0 where test.name === "typed arrays / DataView (Int32)".

I was thinking we wouldn't have to hardcode each test but that doesn't work.

hzoo:
Ok my pr is https://github.com/babel/babel-preset-env/pull/62

data/builtInFeatures.js
yavorsky: Yes, cool! It was rather a hotfix then a long-term solution.

PR #62, opened 2016-12-06
account for multiple features

hzoo:


existentialism:
:+1:

I was doing the same thing locally, but was trying to find a "cleaner" solution before getting distracted. Think this is fine in the interim?

hzoo:
Everything about this project isn't clean 😄  but yeah we can improve

PR #64, opened 2016-12-08
Use array features for symbol and array.iterator

existentialism:


hzoo:
Ok the tests failed due to symbol but we should try to incorporate https://github.com/babel/babel-preset-env/pull/30 instead - i'l do it

PR #79, opened 2016-12-12
Update eslint config to align with other babel projects

baer:
This PR does 2 things:

* Update the project to depend on the latest version of eslint related packages
* Update the eslint configuration to be consistent with the other babel projects ([babili](https://github.com/babel/babili), [babylon](https://github.com/babel/babylon)) in that there is an `eslintrc` file at the root to match the `eslintignore`

__Note__: No linting rules were changed in this PR.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,7 @@
+{
+  "extends": "babel",
+  "parserOptions": {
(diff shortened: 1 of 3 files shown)


hzoo:
Cool 👍

PR #80, opened 2016-12-12
Update pathnames to avoid uppercase

baer:
This PR updates all paths to be kebab case. Not only is this consistent with other babel projects but this helps prevent annoying git errors related to switching between UNIX (case sensitive file system) and MacOS/Windows (case-insensitive file system).

Side note: Not sure if this ever comes up in PRs but if it does, it may be worth adding `eslint-plugin-filenames` to the babel official eslint config.


hzoo:
added in https://github.com/babel/babel-preset-env/commit/03947da35ddfaa02651085ab2e5406cd340c7f00 👍

PR #82, opened 2016-12-12
Update linting rules to cover all js

baer:
This PR updates the linting to cover all files including those in the data directory since they appear to be (for now) hand generated. In the future, it may be good to just add the `/data` directory to the `.eslintignore`.


hzoo:
Awesome

PR #83, opened 2016-12-12
Drop support for node 0.10 and 0.12

baer:
No other babel projects still support the 0.10 / 0.12 Node versions. Lets save a few CPU cycles for travis and drop them here too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,6 @@ node_js:
 - '7'
 - '6'
 - '4'
-- '0.12'
-- '0.10'
 script:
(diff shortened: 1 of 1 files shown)


hzoo:
FYI Babel itself does support node 0.10/0.12 still (check circle/travis)

baer:
Oh, thats my mistake! I thought I remembered in the Babel Slack that it had been dropped. Thanks for catching this!

hzoo:
(but we should definetely figure out when and how to do it)

Ref https://github.com/babel/babel/issues/4315

baer:
I just subscribed to the issue. I'll make an issue in babel-preset-env that
we can use to track it

PR #87, opened 2016-12-12
Cleanup `lib` before rebuilding

baer:
After the kebab case rename I realized that the lib dir wasn't getting cleaned on rebuild. As compiled / data files change it would probably be good to make sure we start fresh with each build.


hzoo:
Ah true we do a similar thing in Babel https://github.com/babel/babel/blob/master/Makefile#L18

existentialism:
Nice catch!

Issue #60, opened 2016-12-06
Add blacklist for transform plugins


johnthecat:
Hi!
I've found some problem with usage of this preset.
My config looks like this:
```javascript
presets: [
    ['env', {
        targets: {
            ie: 11,
            chrome: 47, //Sputnik browser,
            firefox: 'last 2 versions'
        },
        loose: true
    }]
],
plugins: [
    'fast-async'
]
```
And it's doesn't work, because preset tries to transform async functions with `babel-plugin-transform-async-to-generator` instead of `fast-async`. 
It looks strange to me, I think, user should have ability to ignore some plugins.
In my opinion, it should look like this:
```javascript
presets: [
    ['env', {
        blacklist: ['babel-plugin-transform-async-to-generator']
    }]
]
```

yavorsky:
Nice idea! If we already have a `whitelist` option, why we couldn't add `blacklist`?

Moreover, *babel-preset-env* is just a preset. It accepts most options that are common to other presets (loose, modules, etc). Maybe blacklist and whitelist options would be useful for other presets also?

cc @hzoo

hzoo:
some background: presets didn't have options until recently since Sebastian didn't want it to be too complicated and it would just be an array of plugins. If you wanted to customize a preset you would just make a new one yourself. It seemed fine at the time.

It was when we were seeing lots of es2015-loose, es2015-modules, es2015-webpack, etc that it was useful for us to add preset options and then also add a loose/modules option to es2015. I guess we are being conservative in that we should determine whether it's really necessary instead of adding more and more features/options.

It's better to allow users to experiment and all that before we decide to just add stuff. But yes whitelist/blacklist both are interesting.

johnthecat:
Hi, @hzoo ! 
About 'fast-async' : it makes collision with 'async-to-generator' and result code looks ugly. 
About blacklist: after researching source code I found that it's easy to add blacklist into options, I will make PR soon. Thank for fast response!

hzoo:
Cool, I can add this as well @johnthecat.

@MatAtBread feel like I didn't see this plugin before! Is is the same as https://github.com/babel/kneden? @marten-de-vries made that and we added it to the babel org if anyone wanted to help maintain it since it's not complete in functionality yet. Sounds really cool though and maybe we can merge the projects?

hzoo:
I'l just do this now @johnthecat, thanks for offering though

MatAtBread:
[...]

PR #89, opened 2016-12-12
add exclude option, rename whitelist to include

hzoo:
Fixes https://github.com/babel/babel-preset-env/issues/60

cc @johnthecat, @yavorsky


test/fixtures/preset-options/blacklist-regenerator/expected.js
```diff
@@ -0,0 +1,85 @@
+import "core-js/modules/es6.typed.data-view";
```
hzoo: this is long but the point is that it's not importing regenerator-runtime

src/index.js
```diff
@@ -156,25 +157,26 @@ export const validateModulesOption = (modulesOpt = "commonjs") => {
...
+    throw new Error(`The '${type}' option must be an Array<string> of plugins`);
   }
 
-  return whitelistOpt;
-};
+  for (let opt of opts) {
+    if ([
+      ...Object.keys(pluginFeatures),
+      ...Object.values(MODULE_TRANSFORMATIONS)
+    ].indexOf(opt) === -1) {
```
baer: http://node.green/#Array-prototype-includes for a slightly nicer syntax

src/index.js
```diff
@@ -156,25 +157,26 @@ export const validateModulesOption = (modulesOpt = "commonjs") => {
...
-      was passed "${whitelistOpt}" instead
-    `);
+function validatePluginsOption(opts = [], type) {
+  if (!Array.isArray(opts)) {
+    throw new Error(`The '${type}' option must be an Array<string> of plugins`);
   }
 
-  return whitelistOpt;
-};
+  for (let opt of opts) {
```
baer: Actually, this whole thing could be:

```js
difference([
  ...Object.keys(pluginFeatures),
  ...Object.values(MODULE_TRANSFORMATIONS)
], opts);
```

Which would have the added value of being able to call out multiple errors at once rather than bailing at the first one.

https://lodash.com/docs/4.17.2#difference

src/index.js
```diff
@@ -226,7 +229,10 @@ export default function buildPreset(context, opts = {}) {
     }
   }
 
-  const allTransformations = [...transformations, ...whitelist];
+  const allTransformations = transformations
+  .filter((plugin) => blacklist.indexOf(plugin) === -1)
```
baer: Same on the `Array.prototype.contains`

src/index.js
Kovensky: `new Set([...]).has(opt)` 😆

hzoo: This is in the transform so we can't use new features unless we want to use transform-runtime again or we drop old node

src/index.js
hzoo: Sounds like the order should be switched so that `opts` is first in order to return the extra opts

src/index.js
hzoo: And you probably mean `includes` here 😛

hzoo:
Was simple enough to just append to an array just doing that

mikesherov:
👍

README.md
```diff
@@ -113,15 +113,29 @@ Can be `false` to not transform modules.
...
 `console.log` out the targets and plugins being used as well as the version specified in `/data/plugins.json`.
 
-### `whitelist`: `Array<string>`
+### `includes`: `Array<string>`
+
+> `whitelist` is deprecated and will be removed in the next major in favor of this.
 
 Defaults to `[]`
-Enable a whitelist of plugins to always include.
+Enable a list of plugins to always include.
```
existentialism: Nit, but maybe simply:

> An array of plugins to always include.

(To mirror exclude)

README.md
```diff
@@ -113,15 +113,29 @@ Can be `false` to not transform modules.
...
+
+Valid options include any of the [babel plugins](/data/plugin-features.js) or (built-ins)(/data/built-in-features.js) such as `transform-es2015-arrow-functions` or `map`, `set`, `object.assign`.
```
existentialism: Formatting around built-ins link (should be brackets?)

README.md
```diff
@@ -113,15 +113,29 @@ Can be `false` to not transform modules.
...
 
 Useful if there is a bug in a native implementation, or a combination of a non-supported feature + a supported one doesn't  work.
 
 Ex: Node 4 supports native classes but not spread.
 
+### `excludes`: `Array<string>`
+
+Defaults to `[]`
+An array of plugins to always exclude/remove.
+The possible options are the same as for the `includes` option.
```
existentialism: Nit:

> The possible options are the same as the `includes` option.

src/index.js
```diff
@@ -11,6 +12,12 @@ export const MODULE_TRANSFORMATIONS = {
   "umd": "transform-es2015-modules-umd"
 };
 
+export const validIncludeExcludes = [
+  ...Object.keys(pluginFeatures),
+  ...Object.values(MODULE_TRANSFORMATIONS),
+  ...Object.keys(builtInsList).slice(4)
```
existentialism: For future ref, it might be valuable to add a note explaining the slice here

hzoo: Yeah.. it shouldn't even be a part of it really

hzoo:
Oops include/exclude sounds so much better than includes/excludes now that I see it.

hzoo:
- [ ] Ok forgot to make it error if you have both something in both include/exclude

README.md
```diff
@@ -285,6 +299,23 @@ syntax-trailing-function-commas {}
 transform-es2015-arrow-functions {}
 ```
 
+### Example with `exclude`
```
yavorsky: Maybe combine it with `include` examle to prevent too large README?

hzoo:
alright!

yavorsky:
Awesome!

hzoo:
Will merge/release after lunch

src/index.js
```diff
@@ -11,6 +12,12 @@ export const MODULE_TRANSFORMATIONS = {
   "umd": "transform-es2015-modules-umd"
 };
 
+export const validIncludesAndExcludes = [
+  ...Object.keys(pluginFeatures),
+  ...Object.values(MODULE_TRANSFORMATIONS),
+  ...Object.keys(builtInsList).slice(4) // remove the `es6.`
```
maslianok: @hzoo You've probably wanted to write
```javascript
...Object.keys(builtInsList).map(builtIn => builtIn.slice(4))
```

hzoo: Great catch! will fix :+1:

Issue #72, opened 2016-12-10
Exec test for useBuiltIns


hzoo:
So we don't require a module that isn't even available via #68 and #71

and exec test: https://github.com/babel/babel/blob/master/CONTRIBUTING.md#writing-tests

PR #95, opened 2016-12-14
test actual requires from useBuiltIns

hzoo:
Depends on https://github.com/babel/babel/pull/4999
Fixes https://github.com/babel/babel-preset-env/issues/72


test/fixtures/preset-options/use-builtins-all-exec/exec.js
```diff
@@ -0,0 +1 @@
+import "babel-polyfill";
```
hzoo: Because this is an exec.js test, it will actually run all the requires so it would fail if it wasn't found?

PR #107, opened 2016-12-22
Show error if target version is not a number

### LINKED ISSUE #106: Async functions with parameter destructuring don't work on Node 4.x
```
  var _ref = _asyncToGenerator(function* ({ bar }) {});
                                          ^

SyntaxError: Unexpected token {
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:373:25)
    at loader (/private/tmp/foo/node_modules/babel-cli/node_modules/babel-register/lib/node.js:144:5)
    at Object.require.extensions.(anonymous function) [as .js] (/private/tmp/foo/node_modules/babel-cli/node_modules/babel-register/lib/node.js:154:7)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Function.Module.runMain (module.js:441:10)
    at /private/tmp/foo/node_modules/babel-cli/lib/_babel-node.js:159:24
    at Object.<anonymous> (/private/tmp/foo/node_modules/babel-cli/lib/_babel-node.js:160:7)
    at Module._compile (module.js:409:26)
```

Self-contained test: https://github.com/djanowski/babel-preset-env-async-destructuring

existentialism:
Fixes #106 for now since the docs/examples all state number.

We can follow up with string support later.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -43,11 +43,12 @@ export const isPluginRequired = (supportedEnvironments, plugin) => {
       const lowestImplementedVersion = plugin[environment];
       const lowestTargetedVersion = supportedEnvironments[environment];
 
-      if (lowestTargetedVersion < lowestImplementedVersion) {
-        return true;
+      if (typeof lowestTargetedVersion === "string") {
+        throw new Error(`Target version must be a number, 
+          '${lowestTargetedVersion}' was given for '${environment}'`);
       }
 
-      return false;
+      return lowestTargetedVersion < lowestImplementedVersion;
     });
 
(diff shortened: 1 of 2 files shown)


milesj:
Why not just wrap in a `parseFloat()`? It would solve all use cases.

```
parseFloat('1.2') -> 1.2
parseFloat('1.2.3') -> 1.2
```

existentialism:
@milesj don't disagree, and was actually the original version of this PR!

I think @hzoo wanted to get a quick bug fix out before the holidays, and address it alongside some additional refactoring/cleanup shortly after. I'll let him chime in tho 😄

Issue #91, opened 2016-12-13
Add a "target preset" option?


hzoo:
```js
{
  "presets": [
    "env", {
      "targetPreset": "experimental-stage-3" // default to latest like it is currently,
      "targets": {
        "browsers": ["last 2 versions"]
      },
      "exclude": ["transform-regenerator"],
    }
  ] 
}
```

TLDR - figure out how we should include stage-3 plugins.

---

Oh man naming is hard: basically there's a difference between the "target javascript" you want to write in. What is currently "targets" is the list of environments you support (why didn't I just call it "environments" when the preset is preset-env lol 🤔 ).

The values of this option could potentially be any year, latest, or stage 3 (basically the current official presets that we have). In reality I think the only ones we care about would be `latest` and `stage 3`. The reason why [we didn't add stage-3 earlier](https://github.com/babel/babel-preset-env#support-all-plugins-in-babel-that-are-considered-latest) was because I kept thinking about the latest spec rather than forgetting that at stage-3 browser vendors will start implementations. Many projects/companies choose to target stage-3 (and if you are doing lower without codemods or serious consideration of the consequences then there will be issues and that's why we have been discussing https://github.com/babel/babel/issues/4914 and others).

hzoo:
For this we can just do https://github.com/babel/babel-preset-env/pull/113/files but explicitly include the plugins (but only enable them if the relevant option is specified). Maybe the option can just be `stage3: true` - otherwise we go back to babel 5 to have a `stage: 1` option although we just need to be clearer about what that actually means to users (stage 0 meaning it's just an idea that can change or be removed at any point).

rtsao:
@hzoo My initial reason for not manually specifying the plugins in #113 was I didn't like the idea of having two sources of truth regarding which plugins were used in which preset. If this were abstracted out somehow, the list wouldn't need to be duplicated.

PR #113, opened 2016-12-31
Add target presets option

rtsao:
Fixes https://github.com/babel/babel-preset-env/issues/91


rtsao:
Travis tests are passing on Node versions `4` and `7` but failing on `0.10`, `0.12`, and `6`, which I find puzzling. Locally, tests are passing on Node 6 for me.

It seems the the failing cases are caused by inconsistency of `"use strict"` being inserted. For example:

```diff
+'use strict';
+
const a = Math.pow(1, 2);
 
const b = () => 'foo';
```

hzoo:
Yeah this works but I think we can just simply add the stage-3 plugins manually as deps rather than use the preset itself - there were also thoughts of removing presets or renaming since it changed so much so not sure.

rtsao:
Makes sense to me.

I had the idea of allowing this work with arbitrary presets and prevent this package from needing to care about which plugins belong to which preset. However, this is probably more complex than it's worth.

hzoo:
@rtsao I think we can totally do this just adding the plugins individually if you still want to do it

PR #114, opened 2017-01-01
[WIP] Add node's "engines" option

yavorsky:
**Motivation:**

In many cases, targets' `node` option is matching engines from project's `package.json`. Devs have to duplicate this value. In cases with ranges like `> 0.12` they must determine next node.js version and specify it in env-preset's `targets`.
`"current"` or `true` option is not fully covering preset's mission. We could have last node, but  maintain older versions.

So for ex:
*babelrc*

```
{
  presets: [
    [ "env", {
      "targets": {
        "node": "engines",
        "browsers": [ "chrome 54" ]
      }
    }]
  ]
}
```
*package.json*
```
  "engines" : { "node" : ">=0.12" }
```
will use targets:
```
{
  node: 0.12,
  chrome: 54
}
```


yavorsky:
Also, we could use `filterSatisfiedVersions`, `getVersionsFromList ` and `desemverify` from there to parse versions including ranges if target version is not a number.
```
targets {
  "node": ">=5.0",
  "chrome": "^1.1"
}
```
https://github.com/babel/babel-preset-env/pull/107
https://github.com/babel/babel-preset-env/issues/106

existentialism:
@yavorsky I was thinking the same thing, in fact had a rough branch using `semver` dep as well!

On an aside, should we move adding the yarn lock file to a separate PR (similar to https://github.com/babel/babel-preset-env/pull/59) with CI runs with and without it?

test/index.js
```diff
@@ -25,6 +25,22 @@ describe("babel-preset-env", () => {
...
+    it("should return the current node version with option 'engines'", function() {
+      assert.deepEqual(babelPresetEnv.getTargets({
+        node: "engines"
+      }), {
+        node: 0.12
+      });
+    });
+  });
+
+  describe("getLowestFromSemverValue", () => {
```
existentialism: Nit, but may want a test `*` returns null?

yavorsky:
Oh, lock.file, always forget 😤. I think we must add it to `.gitignore`.

test/index.js
yavorsky: Good point! Need to cover `*` as well. Thanks!

yavorsky:
Added:

- envEngines support: 
  - Getting value from: `BABEL_ENV` -> `NODE_ENV` -> `'development'`. 
  - If no `devEngines` found with development environment - trying to access `engines`.
- Remove yarn.lock, add it to .gitignore.
- Update tests:
  - Cover '*' engines value
  - Cover engines value with wrong format.
  - Cover development/production environments.

package.json
```diff
@@ -7,6 +7,12 @@
   "license": "MIT",
   "repository": "https://github.com/babel/babel-preset-env",
   "main": "lib/index.js",
+  "engines" : { 
+    "node" : ">=0.12.0"
```
hzoo: 0.10 here?

package.json
```diff
@@ -7,6 +7,12 @@
...
+  },
+  "devEngines" : { 
+    "node" : ">= 4.x"
```
hzoo: we can do 6 here (I dont remember)

sindresorhus:
Great improvement, but as you already commented, I'd like to see it as the default. Having the `engine` field defined means you're targeting a specific minimum Node.js version, so there's no point in having to also define that you want to use it here.

.gitignore
```diff
@@ -2,4 +2,5 @@ node_modules
 lib
 .DS_Store
 *.log
-.vscode
\ No newline at end of file
+.vscode
+yarn.lock
```
abouthiroppy: Why do you ignore `yarn.lock`?

yavorsky: @abouthiroppy yarn support will be added with separate PR.

yavorsky: So, to prevent commiting it every time unless it will be added, let's place it to .gitignore.

package.json
yavorsky: Yes, updated

package.json
yavorsky: 👍

hzoo:
I guess the only issue is that the node version is different than browser version if you are using it for front-end code. I can build with node 6 but still support ie?

Not sure how we could now to do it automatically since we don't know if it's a node project. However we could make a wrapper for this that assumes node?

yavorsky:
@hzoo Maybe consider engines if no `targets` was specified?

hzoo:
Same issue - you can not specify targets and you still might be compiling for front-end code

sindresorhus:
I incorrectly assumed `node: true` was the default.

---

> ask for https://github.com/nkt/babel-preset-node Would use the minimum of engines + current node? and have no config

👍

yavorsky:
But if no engines/devEngines was found should we use lowest possible node version or current ?

hzoo:
I'm not sure - I would probably use the current node version if you are using the new pkg + no engines, could use 

and maybe we should move out babel-node into https://www.npmjs.com/package/babel-node @jonnyburger.

JonnyBurger:
I can make the `babel-node` npm name available for you if you request.

yavorsky:
For now, `engines: true` means **use current node's version**. Changing it to **use version from engines** might be confusing. Maybe add warning for users who are using `node: true` currently and expecting current node support? 
As for me, `true` option is a little blurry at all. We already have different cases for using `node` option - read from engines, use current, use specified version. And `true` option looks out of the ordinary and not intuitive.

nkt:
Hey guys, i've noticed you asked me for `babel-preset-node`, I'm ready to transfer this package name to community.

danielbayley:
I'm currently finishing off a kind of wrapper API around this preset and the latest CoffeeScript [v2] (for easy/zero config, to hook up source maps properly and because I want it for a bunch of Atom packages…)  and had started to impliment my own check for `engines` when it occured to me it would probably be better as a PR on here, then I found this one… Point being, I'm keen to see this merged, so anything I can do to help I'm in… what, [if any] current blockers are there?

One thing to mention also, is that in the case of Atom packages, `engines` will usually have an `atom` property specifying it's own version (which really just corresponds to a particular version of `electron`), so I'm wondering how to handle this… would that need to be filtered out or just be ignored by this preset anyway? It would be cool if it was automatically resolved to the correct version of `electron`, but that does _feel_ kind of out of scope here…

ELLIOTTCABLE:
I'm with @yavorsky, here — I think `node: true` should be deprecated, in favour of an explicit (and required) string.

k15a:
Hey,

that's an awesome feature. 🎉
Could we also add support for a new option which would take the engines object directly?
```json
{
  "presets": [
    [ "env", {
      "targets": {
        "node": { "node" : ">=0.12" },
        "browsers": [ "chrome 54" ]
      }
    }]
  ]
}
```

That would allow people to get the information from another source and pass it directly to the preset. I think that wouldn't be too hard because we can basically reuse the current logic.

yavorsky:
@k15a We can avoid `node` field duplicating and pass semver value directly.

```json
{
  "presets": [
    [ "env", {
      "targets": {
        "node": ">=0.12"
      }
    }]
  ]
}
```

k15a:
@yavorsky Does the node field also accept ranges? I thought it currently only accepts version numbers.

yavorsky:
@k15a Not yet. When we will merge this PR it would be ease to handle semver values.

sheerun:
Maybe @yavorsky @hzoo Maybe it could first respect engines in `publishConfig`. It would allow for different engines for development and different for publication.

hzoo:
Thanks for the PR! 

PR #121, opened 2017-01-05
Explicitly resolve lowest browser version

### LINKED ISSUE #119: ios version issue
because of browserlist not returning version in a reliable order (https://github.com/ai/browserslist/issues/99 ) if the version is a 'range (`10-10.1` for example) babel-preset-env doesn not handle correctly ios version

if you use the following targets configuration
```javascript
targets: {
  browsers: [
    '> 1%',
    'last 3 versions',
    'ios >= 6',
    'not ie < 10'
  ]
}
``` 
you get
```javascript
// Using targets:
{
  "chrome": 49,
  "edge": 12,
  "firefox": 48,
  "ie": 9,
  "ios": 10,
  "safari": 9
}
```

with ios being 10 instead of the expected 6.

this happens because babel-preset-env strongly relies on the list being ordered and just takes the last returned version.
```javascript
return browsers.reduce(function (all, browser) {
  var _browser$split = browser.split(" "),
      browserName = _browser$split[0],
      browserVersion = _browser$split[1];

  if (browserName in browserNameMap) {
    all[browserNameMap[browserName]] = parseInt(browserVersion);
  }
  return all;
}, {});
```

maybe would be a good idea to do not rely on the order and just explicitly get the smallest version :

```javascript
return browsers.reduce(function (all, browser) {
  var _browser$split = browser.split(" "),
      browserName = _browser$split[0],
      browserVersion = _browser$split[1];

  if (browserName in browserNameMap) {
    all[browserNameMap[browserName]] = Math.min(all[browserNameMap[browserName]] || Infinity, parseInt(browserVersion));
  }
  return all;
}, {});
```

brokenmass:
stop relying on browserlist returning a list sorted by browser version
fix #119


existentialism:
Nit, but can we dasherize the fixture names (`ios-6`, `ios-10`, etc)?

brokenmass:
@existentialism : oops my bad, fixed.

brokenmass:
modified as requested

yavorsky:
@brokenmass Looks good! Also the tests will fail if we add some shims in the future. But we haven't any features < ios 9, so using builtIns seems to be the only way if we want to test ios 6. 
Thanks 👍

hzoo:
We should have a test with multiple browserlist entries like the original issue right? I don't see how the tests cover the changes in code atm since it seems to just test iOS (which is good too)

brokenmass:
As the problem was caused by the getLowestVersions algorithm the query `ios >= 6` is enough to generate the issue, as it previously resolved as `{ios: 10}` because browserlist returned an unordered list.

PR #124, opened 2017-01-08
Include yarn.lock and update CI

existentialism:
Another option is extending the matrix to include testing yarn on _all_ supported versions of node (>4)


hzoo:
I'm ok with that (testing yarn on supported versions and npm on others)?

Side effect of rollup on babylon meant that yarn did that https://github.com/babel/babylon/blob/008a29cb5fdb8309b42a921f7a2fa4110c223f01/.travis.yml#L13

existentialism:
Updated:

image

package.json
```diff
@@ -54,7 +54,7 @@
     "babel-plugin-transform-flow-strip-types": "^6.8.0",
     "babel-preset-es2015": "^6.14.0",
     "babel-register": "^6.14.0",
-    "compat-table": "github:kangax/compat-table#e732718eab42c6c83a364450f456474638d31f94",
+    "compat-table": "kangax/compat-table#e732718eab42c6c83a364450f456474638d31f94",
```
hzoo: can we do this now? @chicoxyzzy

chicoxyzzy: It doesn't change commit hash so I assume it's safe change

PR #128, opened 2017-01-09
Transformations before logs

yavorsky:
1. Log transformations after `excluded/included` processing. Fixes https://github.com/babel/babel-preset-env/issues/127.
2. Prevent a second iteration by replacing filtration from 2 to 1 call:

*before*
```
items
  .filter((builtInName) => isPluginRequired(targets, pluginsList))
  .filter((plugin) => exclusions.indexOf(plugin) === -1)
```
*now*
```
  const filterBuiltIns = filterItem.bind(null, targets, exclusions, pluginsList)
  items.filter(filterBuiltIns)
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -235,7 +235,7 @@ const filterItem = (targets, exclusions, list, item) => {
   if (isDefault) return notExcluded;
   const isRequired = isPluginRequired(targets, list[item]);
   return isRequired && notExcluded;
-}
+};
 
 export default function buildPreset(context, opts = {}) {
   const loose = validateLooseOption(opts.loose);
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -266,26 +278,22 @@ export default function buildPreset(context, opts = {}) {
     transformations.forEach((transform) => {
       logPlugin(transform, targets, pluginList);
     });
-    console.log("\nUsing polyfills:");
     if (useBuiltIns && polyfills.length) {
+      console.log("\nUsing polyfills:");
```
hzoo: 👍  noticed that too

hzoo:
Would also be nice to set up a test for `debug: true` sometime

PR #131, opened 2017-01-09
Remove unnecessary extension

roman-yakobnyuk:
Removed unnecessary `.js` extension and fixed tests.

### DIFF
--- a/data/built-in-features.js
+++ b/data/built-in-features.js
@@ -120,7 +120,7 @@ const es2015 = {
 };
 
 const es2016 = {
-  "es7.array.includes.js": "Array.prototype.includes",
+  "es7.array.includes": "Array.prototype.includes",
 };
 
 const es2017 = {
--- a/data/built-ins.json
+++ b/data/built-ins.json
@@ -658,7 +658,7 @@
     "ios": 10,
     "opera": 25
   },
-  "es7.array.includes.js": {
+  "es7.array.includes": {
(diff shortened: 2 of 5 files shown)


hzoo:
Ok this looks great!

I just realized I gave the wrong instructions though so we need to change https://github.com/babel/babel-preset-env/blob/master/data/built-in-features.js#L123 too

```js
const es2016 = {
  "es7.array.includes.js": "Array.prototype.includes",
};
```

roman-yakobnyuk:
Updated the branch to include this change too, all tests seem fine too.

existentialism:
Nice work @roman-yakobnyuk!

hzoo:
yeah `data/built-in-features.js` is the data and the other file is generated but didn't want to make it too complicated - will merge once CI passes.

hzoo:
Oh yeah a tip: you can put `Fixes #issue-number-here` in the description/commit message and it will automatically close the corresponding issue.

roman-yakobnyuk:
Will keep that in mind for future, thanks!

skattyadz:
Just so you're aware, this PR caused our config to break. It's easy to fix but I wonder if this could have been a minor server bump

hzoo:
I had no idea @skattyadz, I'm not sure why I had it originally and so I figured it was unnecessary (thus no reason for just a patch bump). 

Can you explain the error and all that?

PR #135, opened 2017-01-10
Update compat-table, build data

hzoo:


data/built-ins.json
```diff
@@ -666,34 +673,23 @@
...
-  },
-  "es7.object.get-own-property-descriptors": {
-    "chrome": 54,
-    "firefox": 50,
-    "node": 7,
-    "opera": 41
-  },
+  "es7.object.values": {},
+  "es7.object.entries": {},
+  "es7.object.get-own-property-descriptors": {},
```
yavorsky: Why we missed these values? 😶

chicoxyzzy: because of this https://github.com/kangax/compat-table/commit/5d08a40acbd5744d61e1bb841928a11380b87d6f

hzoo: Oh I didn't push those for some reason (did it before)

data/built-ins.json
```diff
@@ -35,8 +35,7 @@
     "edge": 12,
     "firefox": 4,
     "safari": 6,
-    "node": 0.12,
-    "ios": 10
```
hzoo: need to fix these

hzoo:
Updated, only a few went down (ios10 to ios9)

data/built-ins.json
```diff
@@ -100,26 +100,31 @@
   },
   "es6.map": {
     "chrome": 51,
+    "firefox": 53,
```
chicoxyzzy: Not sure that iterator closing is such significant feature for Map and other new types. FF can deal with these types since 30/40th versions

hzoo: Yeah its not, haven't found a good way to account for these for us

hzoo: again trying to be correct first/conservative and then we can optimize like in https://github.com/babel/babel-preset-env/pull/30

ljharb: Every single feature and observable thing is significant - if something isn't 100% compliant, it's broken, and it needs to be patched/polyfilled/worked around.

PR #138, opened 2017-01-13
README: Update `debug: true` example.

yavorsky:
According to [last changes](https://github.com/babel/babel-preset-env/pull/109).


README.md
```diff
@@ -286,16 +286,24 @@ exports.A = A;
 ### Example with `debug: true`
 
 ```js
-Using targets: {
-  "node": 6.5
+Using targets:
```
existentialism: Thoughts on including the config used to generate this (similar to the other examples?)

```json
{
  "presets": [
    ["env", {
      "targets": {
        "safari": 10
      },
      "useBuiltIns": true,
      "debug": true
    }]
  ]
}
```

Issue #142, opened 2017-01-15
replace manual electron-to-chromium file with external package


Kilian:
I just created an electron-to-chromium package that provides the same object that you manually keep in https://github.com/babel/babel-preset-env/blob/master/data/electron-to-chromium.js but then automatically generated as suggested by @kevinsawicki here: https://github.com/babel/babel-preset-env/pull/55#discussion_r91378631

Would you consider replacing the file with `require('electron-to-chromium').versions`?

The package is here: https://www.npmjs.com/package/electron-to-chromium
The git repository is here: https://github.com/Kilian/electron-to-chromium

It was created after discussion in the browserlist repository: https://github.com/ai/browserslist/issues/96 Where the main goal for me was to make sure Browserlist understood "Electron >= 1.4", but @ai (rightfully) wanted to keep it out browserlist until it proved popular enough. Given that you already use a mapping file, switching to electron-to-chromium makes more sense for you.

hzoo:
Yep for sure, can make the PR if you'd like as well (I didn't want to make a mapping myself so I asked so this is great)

Kilian:
Will do :)

Kilian:
I'm working on replacing the file with my package, but one of the tests are failing. However, the test in question says that electron 1.0 uses chrome 50. My package and the actual list says it uses chrome 49: https://gh-contractor-zcbenz.s3.amazonaws.com/atom-shell/dist/index.json so i'll be sure to update the tests too :)

Kilian:
Additionally, other tests assumed electron 1.5 === chrome 54, but that's also not a given (yet). I've updated both, PR incoming!

PR #144, opened 2017-01-16
Use external Electron to Chromium library

Kilian:
Per the suggestion in ticket #142 this pull request switches out the local list of electron-to-chromium mappings for a more complete external library.

I had to update a number of tests, because Electron 1.5 is not out yet and it's not guaranteed it will use the assumed chrome version. Additionally, the old data file linked electron 1.0 to chrome 50 but that should have been 49, which I updated in the tests as well.

I updated the external library to allow both string and numerical input, but it always returns a string (because it can return full chromium versions too, which require it) so I `parseInt()` that in the script. I hope that's ok.


test/fixtures/preset-options/electron/options.json
```diff
@@ -2,7 +2,7 @@
   "presets": [
     ["../../../../lib", {
       "targets": {
-        "electron": 1.5
+        "electron": "1.4"
```
yavorsky: Maybe avoid strings unless we will maintain it for all targets?

src/index.js
```diff
@@ -141,7 +121,10 @@ export const getTargets = (targets = {}) => {
 
   // Rewrite Electron versions to their Chrome equivalents
   if (targetOps.electron) {
-    targetOps.chrome = electronVersionToChromeVersion(targetOps.electron);
+    targetOps.chrome = parseInt(electronToChromium(targetOps.electron), 10);
```
yavorsky: If we have for ex:
```js
"targets": {
  "chrome": 50,
  "electron": 1.3 // 52
}
```
looks like it will just rewrite chrome version with higher. So the result will be chrome: **52**, but not **50**.

test/fixtures/preset-options/electron/options.json
Kilian: will do. I did this before updating my library to also handle numbers, so it can be reverted.

src/index.js
Kilian: I'll do a Math.min()

Kilian:
I didn't want to change the way the behaviour worked with regards to how it handles the electron -> chrome conversion so that works the same as it currently does: set the targetOps.chrome and delete targetOps.electron. Logging it as electron and keeping the chrome conversion on a lower level is better suited for a different PR, I think.

src/index.js
yavorsky: @Kilian 👍

Kilian:
Just pushed the requested changes!

src/index.js
```diff
@@ -141,7 +121,18 @@ export const getTargets = (targets = {}) => {
...
   // Rewrite Electron versions to their Chrome equivalents
   if (targetOps.electron) {
-    targetOps.chrome = electronVersionToChromeVersion(targetOps.electron);
+    const electronChromeVersion = parseInt(electronToChromium(targetOps.electron), 10);
+
+    if (!electronChromeVersion) {
+      throw new Error(`Electron version ${targetOps.electron} is either too old or too new`);
+    }
+
+    if (targetOps.chrome) {
```
existentialism: This might be a good test case?

Kilian: you're right! will add it tonight.

Kilian:
@yavorsky I think all the requested changes are put in!
@existentialism two test cases!

...And now it can not be merged back to master :( I'll see how much trouble it is to rebase.

Kilian:
Rebased on latest master and all green, should be good to go now.

src/index.js
```diff
@@ -124,7 +103,18 @@ export const getTargets = (targets = {}) => {
 
   // Rewrite Electron versions to their Chrome equivalents
   if (targetOps.electron) {
-    targetOps.chrome = electronVersionToChromeVersion(targetOps.electron);
+    const electronChromeVersion = parseInt(electronToChromium(targetOps.electron), 10);
```
hzoo: I don't think `parseInt(a, 10);` is necessary anymore in ES5? Can remove later

hzoo:
Thanks for the library + PR!

PR #125, opened 2017-01-09
Extract option normalization into independant file

baer:
This PR moves all preset option validation logic into a separate file. 95% of this PR is simply moving code from the `index.js` => `normalize-options.js` with the exception of the following:

- Error messages were edited for clarity and consistency. They are now of the format `Invalid Option: <problem>` or `Deprecation Warning: <problem>`
- The `invariant` library was pulled in to help make the code a bit more readable.
- With the exception of setting default values, `normalizeOptions` does not mutate or change the input object. Part of what set me down the path of this refactor was unexpectedly finding that `validateIncludeOption` and `validateExcludeOption` not only altered the input but changed the shape.
- `lodash/intersection` was pulled into the plugin bundle for clarity.

Thanks for all the great work on this repo! If you find this change too invasive, I'm happy to try to scale back in places. The impetus for all of this was finding that validation was actually mutating which could be fixed in other ways. Thanks for doing all ya'll do here!

PR #145, opened 2017-01-17
Update yarnfile

baer:
The following changes have been made since the last update to the yarnfile.

```
-    "compat-table": "kangax/compat-table#e732718eab42c6c83a364450f456474638d31f94",
-    "eslint": "^3.3.1",
-    "eslint-config-babel": "^3.0.0",
-    "eslint-plugin-babel": "^4.0.0",
+    "compat-table": "kangax/compat-table#b0cec63ea21f3a7788a8eececcb918de903b7fc5",
+    "eslint": "^3.13.1",
+    "eslint-config-babel": "^5.0.0",
```

If yarn is used as the primary package manager it might be worth doing a `yarn install` in CI to make sure it comes up clean. I'm happy to try to set that up. Anyway, this will get the file up to date so I can get [this guy](https://github.com/babel/babel-preset-env/pull/125) into shape without committing extraneous stuff.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1167,17 +1167,11 @@ date-now@^0.1.4:
   version "0.1.4"
   resolved "https://registry.yarnpkg.com/date-now/-/date-now-0.1.4.tgz#eaf439fd4d4848ad74e5cc7dbef200672b9e345b"
 
-debug@2.2.0, debug@~2.2.0:
+debug@2.2.0, debug@^2.1.1, debug@^2.2.0, debug@~2.2.0:
   version "2.2.0"
   resolved "https://registry.yarnpkg.com/debug/-/debug-2.2.0.tgz#f87057e995b1a1f6ae6a4960664137bc56f039da"
   dependencies:
     ms "0.7.1"
-
-debug@^2.1.1, debug@^2.2.0:
-  version "2.6.0"
-  resolved "https://registry.yarnpkg.com/debug/-/debug-2.6.0.tgz#bc596bcabe7617f11d9fa15361eded5608b8499b"
-  dependencies:
-    ms "0.7.2"
 
 deep-eql@^0.1.3:
   version "0.1.3"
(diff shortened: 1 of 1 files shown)


hzoo:
Should we make sure to do `npm run build-data` after each change? - this is to know if there's a bug etc either with our code or via compat-table

hzoo:
Ok yeah getting that issue again.. why.. (removing chrome, ios, etc)

baer:
I actually had a lot of trouble with Yarn :/... What is the issue you're getting?

hzoo:
The issue is that running `npm run build-data` is causing the data/ to change when it shouldn't, not sure if it's yarn or compat-table update

baer:
I just pulled master and got a clean run of `npm run build-data`. That dependency is pinned to a SHA so I'm not sure why our runs would be different.

hzoo:
Ok cool, maybe yarn version/cache/etc 😓 ?

existentialism:
Same, with both `yarn build-data` and `npm run build-data`

baer:
I just ran both
* `rm -rf node_modules && yarn && npm run build-data`
* `rm -rf node_modules && npm install && npm run build-data`

And I got a clean master both times. That narrows it down to a cache issue (on one of our machines.)

baer:
I get this error every time I do a fresh `yarn` on this project:

```
[2/4] 🚚  Fetching packages...
error An unexpected error occurred: "ENOTEMPTY: directory not empty, rmdir '/Users/ebaer/Library/Caches/Yarn/npm-core-js-2.4.1-4de911e667b0eae9124e34254b53aea6fc618d3e/library/fn'".
info If you think this is a bug, please open a bug report with the information provided in "/Users/ebaer/dev/opensource/babel-preset-env/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

It is related to the pinned version of the compat table. When I remove the package, run `yarn`, put it back in and rerun the cmd it runs just fine. My guess is that this is a corner case that yarn just doesn't handle well. I was able to run:

`npm cache clean && rm -rf node_modules && npm install && npm run build-data`

And still get a clean master which again points to a problem with yarn and very likely with the pinned `compat-table`.

hzoo:
Oh I just got that error ^

baer:
Can it be pinned to a regular version? Why use a SHA here?

hzoo:
There are no regular versions since it's not on npm (that's a wip)

PR #152, opened 2017-01-23
Update yarn lockfile

existentialism:
Missed this in https://github.com/babel/babel-preset-env/pull/144


hzoo:
Should setup https://github.com/devtools-html/debugger.html/blob/master/bin/update-yarn-lock-file?

PR #156, opened 2017-01-27
Add tests for debug output

### LINKED ISSUE #133: Add tests for debug output
Now that we've fixed a few bugs in debug output (h/t to @yavorsky), it would be great to add test(s) to make sure we don't break it later.

We definitely want to ensure that:

- Output logs _after_ exclude/include processing is done (https://github.com/babel/babel-preset-env/pull/128)
- Targets include parsed browsers (https://github.com/babel/babel-preset-env/pull/109).

We should be able to test output to stdout for this, similar to [babel-cli](https://github.com/babel/babel/tree/master/packages/babel-cli/test).

existentialism:
Fixes #133.

Based heavily on babel-cli tests, we can follow up with maybe adding support for `stdout` tests either as a new package or as part of `transform-fixture-test-runner`.

**EDIT:** I just remembered that the exec tests added in https://github.com/babel/babel-preset-env/pull/95 only run npm 3, which masked the breakage on node versions that didn't have `Array.prototype.includes` added by https://github.com/babel/babel-preset-env/pull/125.


test/debug-fixtures.js
```diff
@@ -0,0 +1,111 @@
...
+      opts.stdout = helper.readFile(stdoutLoc);
+    }
+
+    const optionsLoc = path.join(testLoc, "config.json");
+
+    if (!fs.existsSync(optionsLoc)) {
+      throw new Error(`Debug test '${testName}' is missing a config.json`);
+    }
+
+    const config = merge({ "debug": true }, require(optionsLoc));
```
yavorsky: Maybe add debug option directly to the config? Because of we actually expect logs and it will just give us more clarity.

existentialism: My thinking was that there's likely never a case when testing debug that we don't want this flag to be set, but I'm fine if we want to be super explicit with a full `.babelrc` file

Issue #149, opened 2017-01-21
Question: Supports `browserslist` config on package.json?


amilajack:
Does this plugin allow using the `browserslist` property in the package.json?

yavorsky:
@amilajack Yes! I'm going to implement this tomorrow!
Today I've made the presentation for a small meetup and thought - "why we are still didn't implement this?" on a slide about common config for preset-env/autoprefixer/eslint-compat. 🤔
We just need to update autoprefixer to ^1.5 and ensure that browserslist doesn't return default browsers if we are not setting up browsers in targets.

amilajack:
🎉🎉🎉 Would play very nicely with [eslint-plugin-compat](http://github.com/amilajack/eslint-plugin-compat)

![demo of plugin usage](https://raw.githubusercontent.com/amilajack/eslint-plugin-compat/master/img/eslint-plugin-compat-demo.gif)

ai:
@yavorsky why you don’t want to use default browsers set if browsers was not specified?

Browserslist default is a good set. Also using default set is a Autoprefixer behaviour right now. We need to change it too if there is some problem with it.

yavorsky:
@ai the idea is to use babel-preset-env without any options like [babel-preset-latest](https://babeljs.io/docs/plugins/preset-latest/). It must be easy to switch from latest preset or just 2015 to env without risks and losses. Also, users need to remember the set of default browsers to know what they actually support.
As I see, with default set, some builtIns won't be included so it might be confusing for people who are not using any option and expecting all things to include.

ai:
@yavorsky OK, I understand logic. But sad that in result Autoprefixer and `babel-preset-env` will be different :(.

Also in my experience users should not set browsers. Most of users make a mistake in setting browsers. A lot of users will set `browserslist: 'last 2 iOS versions'` for mobile website. We already saw a lot of problems with it and `-webkit-` prefix.

This is why main way to use Autoprefixer is to no set browsers. Because default browsers is better for most of users.

What do you think about it? Maybe `babel-preset-env` should not force users to set browsers too, because they will make mistakes in it?

yavorsky:
@ai Yes, I like simplicity too and this preset is a good intention. As autoprefixer/browserslist of course.
But there are some issues:
1. What criteria of selecting default browsers? Why ie 10 or chrome 49 is in default list? 
[...]

Issue #108, opened 2016-12-23
Promote shareable browserslist config


ai:
I think we should promote shareable browsers config instead of `browsers` option. Maybe `browserslist` key in `package.json` will be the best way.

I think most of babel-preset-env users also use Autoprefixer. So for better maintainability they should use shareable browsers config to use same browsers list in Autoprefixer and babel-preset-env. But most of developers doesn’t know about this option and benefits of it.

I [already removed](https://github.com/postcss/autoprefixer/commit/3de19bf91d0b175e3c8c8a42aa12d88d4fa344a3) `browsers` option from all examples in Autoprefixer docs and focused on sharable configs.

yavorsky:
Nice! It might be the first step for [configs generalization](https://github.com/babel/babel-preset-env/pull/51#issuecomment-267148584). 
Moreover we already have browserslist as dependency and could parse config with [parseConfig](https://github.com/ai/browserslist/blob/master/index.js#L360).

I come across cases when people (and my projects too) is using it in bundler's config files, something like:
```
const browsers = ['> 4%', 'ie 11', 'safari 8'];
const envPreset = ["env", {targets: {browsers}}];
```
And use it

TuckerWhitehouse:
Would it be possible to remove the if statement from these lines: https://github.com/babel/babel-preset-env/blob/master/src/index.js#L145-L149.
If the user defines "browsers" in the targets, it will be passed in as before, and if not, browserslist will use it's internal resolution.

ai:
@yavorsky I found that unfortunately current version of `babel-preset-env` don’t use Browserlist config at all :(. It doesn’t execute `browserslist` function if `browsers` options was not passed.

Could we change it? What do you need to change it?

ai:
@yavorsky BTW, this solution is not ideal:

```js
const browsers = ['> 4%', 'ie 11', 'safari 8'];
const envPreset = ["env", {targets: {browsers}}];

...
plugins: [require('autoprefixer')(browsers)]
...
presets:  [envPreset]
```

Not only Autoprefixer and Babel need a browsers list. Stylelint and webpack CSS compression uses (or will use soon) browsers information too. In this case we could not pass browsers to all tools manually.

I agree that config mess is a problem. This is why new Browserslist version supports `package.json` browsers list. Is it a solution for you?

yavorsky:
@ai actually, removing single condition will fix it. 
[...]

PR #161, opened 2017-02-06
Add browserslist config/package.json section support.

yavorsky:
This PR adds support for `browserslist` field in the `package.json` and `browserslist` file (including `path` option to specify entry point for config file searching).

Starting from v1.5, browserslist supports `package.json` and browserslist config even earlier.

As @ai mentioned in https://github.com/babel/babel-preset-env/issues/108, most of babel-preset-env users also use [autoprefixer](https://github.com/postcss/autoprefixer).
Also, @amilajack created amazing [eslint-plugin-compat](https://github.com/amilajack/eslint-plugin-compat) which detects browsers to support from browserslist query.

It's really easy and painless to use **same config value for many things**. It allows us simply consider [environments](https://github.com/ai/browserslist#environments), speed up the time to configure a project, unify configs at all.

browserslist doesn't support node.js yet, but it could be added easily in `targets`. [See](https://github.com/babel/babel-preset-env#support-a-target-option-node-current-to-compile-for-the-currently-running-node-version).

I believe that https://github.com/babel/babel-preset-env/pull/114 will come to life and we could use **browserslist query for front-end stuff** and **engines/devEngines for node.js projects**.

A few examples..
*package.json*
```json
"browserslist": "ie 10, > 2%"
```
or 
*browserslist*
```
[production]
last 2 version
ie 9

[development]
last 1 version
```

Note: we had [fruitful discsussion](https://github.com/babel/babel-preset-env/issues/149#issuecomment-274728128) with @ai, @existentialism and @amilajack about the default result for empty browserslist query. As I mentioned before, default browsers is a good idea at all, but for now, we must be sure that config-less usage doesn't break anything, so there are no default browsers as well as before, so **ignoring browserslist at all won't cause some targets exclusion**.


src/index.js
```diff
@@ -117,12 +113,14 @@ export const getTargets = (targets = {}) => {
 
     delete targetOps.electron;
   }
+  browserslist.defaults = objectToBrowserslist(targetOps);
+
+  const browsersQuery = targetOps.browsers;
+  const browserslistOpts = {path: opts.path};
```
existentialism: Should we rename the `browserslist` path option on our end to something more descriptive?

Maybe `browserslistConfigPath`?

yavorsky: @existentialism I was thinking about this but decided we could have common `path` option for all things. For example, while searching package.json for `engines` in the future. And apparently, it would be the same path for package.json in this case.
I agree that we should pay attention to this, but ideally try not to spread new options, but to unify it and make it works in chime.
If we really could have problems with a common path, `browserslistConfigPath ` is good. Or `browserslistPath`, because it's also necessary for `package.json` searching.

darsain:
If this should eventually accept node versions, shouldn't it be named something environment agnostic, like `"envslist"`? It will feel weird putting `node > 4` into a property named `"browserslist"`.

yavorsky:
@darsain As I mentioned above, browserslist doesn't support node.js currently:
https://github.com/ai/browserslist/issues/75

You can specify `targets: {node: 7.4}` in babel-preset-env's options.

src/index.js
existentialism: Ahh, gotcha.

I guess my only real nit with the name "path" is that in a list of options I'd find myself asking "path to _what_?"

How about `configPath`? I dunno. Ultimately, it's not a big issue and am fine with it if @hzoo or anyone else doesn't care :)

existentialism:
@yavorsky woot, https://github.com/babel/babel/pull/4834 merged!

yavorsky:
@existentialism Wow, thanks!🎉 Going to update path option in 2 hours!

yavorsky:
@existentialism Pushed but seems like tests will fail for older babel versions which not provide `dirname` as a 3 argument for presets.

motiz88:
@yavorsky Needs more tests. I've actually started implementing this because coming across your PR, so should be able to port some of the testing work I've done if you want. It'll take me a couple of days though.

motiz88:
At any rate, when the new version of `babel-core` containing babel/babel#4834 gets released, we should depend on it explicitly, and also add a note to the README mentioning this along with the current note on Browserslist.

yavorsky:
@motiz88 It would be cool! What kind of tests you want to implement. Currently, with fixtures it looks for browserslist config in file's directory and checks whether or not config used like a browserslist query. If config was not found - preset uses targets. And if no targets is specified - works like `preset-latest`.
Also, I've added unit-test for `objectToBrowserslist` to parse plain object into string that browserslist understands.

yavorsky:
@motiz88 Yeah. If we want to cover all things we must be sure that current version of babel-core includes babel/babel#4834.

existentialism:
@yavorsky let's retarget to 2.0, we won't have to worry about tests failing as `dirname` would always be present

yavorsky:
@existentialism It already has 2.0 as a base branch. 🙂 Will resolve conflicts in the morning. Also needs to update readme and discuss the main points before merge (like cases to look up browserslist config, behavior when browsers: [] specified, etc. I will summarize all the things after night (from mobile now 🌙).

existentialism:
@yavorsky whoops, was looking at the old package.json and swore it was on master. sounds good, let's get the remaining issues resolved and land this asap!

hzoo:
We should add an entry to the readme about using it

yavorsky:
@hzoo Yes! In progress...

hzoo:
also we can add this to the debug output - like "using browserslist" or something

README.md
```diff
@@ -122,6 +122,14 @@ A query to select browsers (ex: last 2 versions, > 5%) using [browserslist](http
 
 Note, browsers' results are overridden by explicit items from `targets`.
 
+#### Browserslist config file.
```
existentialism: I think it's useful for this section to describe how `browserslist` config interops with `targets`.

For example how it merges (or doesn't, depending on what we decide).

Then, maybe we should move the `browserslist` default behavior into a main section above options since it's a really powerful default.

Thoughts?

existentialism: Maybe something like:

> ## How It Works
>
> ### Browserslist support
>
> By default, `babel-preset-env` will use [`browserslist` config sources](https://github.com/ai/browserslist#queries). This means that you can specify what targets your project supports in a `browserslist` config file or key in your `package.json` file once, and tools like [autoprefixer](https://github.com/postcss/autoprefixer), [eslint-plugin-compat](https://github.com/amilajack/eslint-plugin-compat), and many others will share it.
>
> For example, to enable only the polyfills and plugins needed for a project targeting `Last 2 versions` and `IE10`:
>
> .babelrc
>
> ```json
> {
>   "presets": [
>     ["env", {
>       "useBuiltIns": true
>     }]
>   ]
> }
> ```
>
> browserslist
>
> ```txt
> Last 2 versions
> IE 10
> ```
>
> You can also disable this via the `ignoreBrowserslistConfig` option.

hzoo:
unrelated to this pr but we should just move the examples at the bottom and inline them for each option under `<details>` - can do something like https://github.com/zeit/next.js#built-in-css-support or not

yavorsky:
Ok, after babel-core alpha.7 update the 3-rd argument for the preset is just `process.cwd()` (https://github.com/babel/babel/blob/7.0/packages/babel-core/src/config/build-config-chain.js#L16). (Before it used `filename && path.dirname(filename)`.)

There is a problem because we need to know current file location and use it as an entry point for babelconfig/package.json search. Have discussed with Logan, but looks like it was cache optimization etc.

Same for the [engines PR](https://github.com/babel/babel-preset-env/pull/114).

loganfsmyth:
I can expand here so everyone has the context. The third param of a preset at the moment is `{ dirname }` (on current Babel 6), added in https://github.com/babel/babel/pull/4834 with the intention that configs would be able to resolve relative paths, so if you did

### foo/.babelrc
```
{
  presets: [
    ['my-preset', {config: "./config.json"}],
  ]
}
```
so `dirname` is the folder `foo` which contains the `.babelrc` so that `my-preset` can do `path.resolve(dirname, options.config)` to read the config. The path is likely a parent of the file being compiled, in the case of `.babelrc` files, but it is not the specific folder containing the file.

When used programmatically instead via `babel.transform` or `babel-loader#options` or whatever, Babel 6 passes the directory of the file being compiled, as mentioned above, or the cwd if no filename is known. This leads to several issues like

```
options: {
  presets: [
    './my-custom-preset'
  ],
}
```
where Babel would search for this path relative to each file that was compiled, which makes relative paths essentially useless. or:


```
options: {
  presets: [
    ['my-preset', {config: './config.json'}],
  ],
}
```
is the expectation that `config.json` exists as a sibling of every possible file being run through Babel, or is it relative to the working directory? Now it will essentially behave the same as Node, e.g.
```
['my-preset', {config: fs.readFileSync('./config.json', 'utf8')}],
```
with it relative to cwd.

This also resolves another issue we had with presets/plugin resolution:
```
projects/
  a/index.js
  b/
    node_modules/
      babel-preset-es2015
      a => ../../a   // symlink
    app.js // require('a')
```
if you configure `babel-loader` with
```
options: {
  presets: ['es2015'],
}
```

this structure would break Babel 6 when Webpack tries to process `a/index.js` because it will try to find `es2015` relative to `project/a` which doesn't even have an `node_modules` folder, let alone the preset.

All of these usecases are fixed by ensuring that Babel's arguments are resolved consistently relative to the same directory. Currently that is hard-coded to cwd, but it may be configurable in the future.

That said, none of this is aimed at the usecase of resolving something relative to the file being compiled.

Personally, I question whether that is the right choice anyway, since a node_module shouldn't really have any influence on what browsers the root application wants to target.

chicoxyzzy:
Can I help here somehow?

ai:
Using `process.cwd` will be bad. Different behavior in different tools.

Maybe we can change Babel API since this project is very important for Babel?

kroko:
Hi,

Till now I have used `browserlist` key in `package.json`.

I'd like to split into file. However, insted of `browserslist` use `.browserslistrc` as it would nicely sit along with other config dotfiles such as

```
.babelrc
.eslintignore
.eslintrc.js
.stylelintrc.js
```

The op mentions _This PR adds support for (..) `browserslist` file (..)_.
Will it work with `.browserslistrc` file?

Thanks!

dvakatsiienko:
+1 on this, would be a great feature

yavorsky:
:D
Nice, 2 approves.
Going to resolve conflicts + add the `path` option. Maybe it's temporary, but good for the current state. I belive we need to land it with 2.0.

sudo-suhas:
> So I think we can use process.cwd as an entry point for config searching

> Using process.cwd will be bad. Different behavior in different tools.

Would [app-root-path](https://github.com/inxilpro/node-app-root-path) be better?

yavorsky:
@sudo-suhas `process.cwd` is not the part of babel-preset-env. Babel in the `build-config-chain` module pass it as [`dirname` field](https://github.com/babel/babel/blob/7.0/packages/babel-core/src/config/build-config-chain.js#L30).

chicoxyzzy:
🎉 Can't wait to try new beta with this feature!

damianobarbati:
Is this merged? Can we finally have the following in package.json?
```
"browserslist": [
    "chrome 58",
    "safari 10"
],
"babel": {
    "presets": [
        [
            "env",
            {
                //no targets!!!!! <3
                "modules": false,
                "useBuiltIns": true,
                "debug": true
            }
        ],
        "stage-0",
        "react"
    ]
}
```
And just this in webpack.config.js?
```
rules: [{
    test: /\.(js|mjs)$/i,
    use: 'babel-loader',
}]
```

yavorsky:
@damianobarbati It was merged and you can try it using latest alpha  - `2.0.0-alpha.19`.

damianobarbati:
@yavorsky I just tested this and it works great.

Gotchas for everybody:

1. some error thrown because of new `useBuiltIns` usage (https://github.com/babel/babel-preset-env/tree/v2.0.0-alpha.19#usebuiltins-usage).

2. I still don't get whether is better "usage" or "entry": assuming I'm using webpack, es6 imports and tree shaking which is the best?

3. I'm transpiling modules which own a `jsnext:main` property: problem is this modules are using the old `useBuiltIns:true` having me to bump versions with fix. Question: shouldn't babel transpile the modules using the "main/parent" babel configurations instead of that one? 🤔

hzoo:
@damianobarbati those questions seem unrelated to this issue/PR and about useBuiltIns, can ask in slack or make a discussion issue about it

damianobarbati:
@hzoo correct, moving them where appropriate.
@yavorsky inserted statements like `import "babel-polyfill/core-js/modules/es6.map";` are resulting into errors to me: I checked `node_modules/babel-polyfill` and there's no `core-js` folder thus the errors.
Weird thing is this happens only with `usage` and not for `entry` option.

BTW: is it correct to report this into this commit thread? Or should I open an issue?

existentialism:
What version of babel-polyfill are you using?

See:
https://github.com/babel/babel/tree/7.0/packages/babel-polyfill/src/core-js/modules

damianobarbati:
solved installing `"babel-polyfill": "^7.0.0-alpha.19"`, thanks

PR #175, opened 2017-02-27
Fix hasBeenWarned condition.

yavorsky:
Remove check for static `hasBeenWarned` that always `false`.

### DIFF
--- a/src/normalize-options.js
+++ b/src/normalize-options.js
@@ -4,8 +4,6 @@ import defaultInclude from "./default-includes";
 import moduleTransformations from "./module-transformations";
 import pluginFeatures from "../data/plugin-features";
 
-const hasBeenWarned = false;
-
 const validIncludesAndExcludes = [
   ...Object.keys(pluginFeatures),
   ...Object.keys(moduleTransformations).map((m) => moduleTransformations[m]),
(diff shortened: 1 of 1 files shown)


src/normalize-options.js
```diff
@@ -71,7 +69,7 @@ export const validateModulesOption = (modulesOpt = "commonjs") => {
 
 export default function normalizeOptions(opts) {
   // TODO: remove whitelist in favor of include in next major
-  if (opts.whitelist && !hasBeenWarned) {
+  if (opts.whitelist) {
     console.warn(
```
hzoo: then we should add hasBeenWarned = true in the if block? otherwise it will warn multiple times?

yavorsky: @hzoo yes! I thought it could be called once, but actually not. Updated!

existentialism:
Should we add a test for this?

hzoo:
oh yeah, always

hzoo:
maybe bot should warn - if you change src and no test files changed 😼

yavorsky:
Maybe pre-commit hook? You can't commit unless added some tests 🙂

I'll cover it after lunch (UTC +2).

hzoo:
Nah since I don't want new contributors to not make the pr if they don't have it, can tell them after?

yavorsky:
I was joking :) But the bot that shows changes coverage - great idea!

hzoo:
And we do have codecov set up on babel already

Issue #134, opened 2017-01-10
Make uglify a target


hzoo:
Going to work on this now

If you use minification already (uglify) then targeting later browers and compiling less would be useful because it will syntax error. I thought at the very least we could have an option to only check polyfills with a `targetBuiltInsOnly` option but making `uglify` its own target makes way more sense.

Of course I will be integrating/try https://github.com/babel/babili so that option wouldn't be needed

skhameneh-sabre:
Using Babili, I can't get my node targeted build to even come close to what Uglify produces, some of this has to do with the order of execution in Webpack.  Running Babili after the project is built causes a heap exception (known issue with very large projects).  I've looked at doing AST transforms myself, but the complexity of operations is introducing concerns (properly hoisting variables requires checking if the declaration is a type of Literal).

Using Acorn with Uglify doesn't make any difference and I can't get the Harmony branch of Uglify to work (too many bugs).  Targeting Uglify would be an intermediate patch, but what's really desired is a proper ES6 compatible Uglify.

Still +1 on this request, let me know if you have any other suggestions - maybe some of what I've said is useful for you as well.

hzoo:
also my intent is to do this for uglify 2 (not harmony) and then working on making Babili better

skhameneh-sabre:
Understood, I'm stating this feature is only useful because Uglify doesn't fully support ES6 and Babili has less transforms than Uglify.

On a side note, I started writing code for hoisting variables (w/ Babel AST).  I left off at handling logic for literals/assignment; I can share that if you'd like to take a look at it, I'm sure there are better methods to manipulate the AST than I was using.

OlegLustenko:
Hi, any news?

yavorsky:
@OlegLustenko Hi! I've decided to solve it after your comment and it works ok, but I need to write some tests and to decide what to do with uglify/harmony. Going to send PR in the morning with a fresh mind. (Seems like we have the same timezone ;) )

OlegLustenko:
Thanks :)

PR #178, opened 2017-03-02
Add uglify as a target

yavorsky:
Add uglify as a target. Fixes https://github.com/babel/babel-preset-env/issues/134.
I've not given attention to uglify's harmony brunch because of there no info about what ES6 features they currently support and how stable this support is (seems like not much for now).

Also, as for me, there is no difference for us between uglify and uglify 2. Both haven't es6 support at all. So allowed uglify values are `number`s and `boolean`s. 

**If uglify hasn't es6 support, what's the difference between no targets specifying at all?** 
The difference is in built-ins. For uglify no matter what polyfills you are using - it's just methods and there are no problems to parse them.

Ex:
*.babelrc*
```json
{"presets":
  [
    ["env", {
      "targets": {
        "chrome": 55,
        "uglify": true
      },
      "useBuiltIns": true,
      "modules": false
    }]
  ]
}
```
Will use **all the possible plugins** and **only built-ins based on specified targets** (except uglify).


README.md
```diff
@@ -122,6 +122,12 @@ A query to select browsers (ex: last 2 versions, > 5%) using [browserslist](http
 
 Note, browsers' results are overridden by explicit items from `targets`.
 
+### `targets.uglify`
+
+`number | true`
+
+If you are using minification with uglify then targeting later browsers would throw a syntax error. To prevent it - specify `uglify` option. It will just use all plugins and as result compile fully to ES5, but with `useBuiltIns` it will include only polyfills according to your targets.
```
yavorsky: @existentialism Oh, thanks!!! 😊

existentialism:
@yavorsky minor, but I pushed a debug test when using uglify output (wanted to make sure it shows "uglify: true)

src/index.js
```diff
@@ -94,29 +94,33 @@ const _extends = Object.assign || function (target) {
...
-  const targetOps = _extends({}, targets);
+  const targetOpts = _extends({}, targets);
 
-  if (targetOps.node === true || targetOps.node === "current") {
-    targetOps.node = getCurrentNodeVersion();
+  if (targetOpts.node === true || targetOpts.node === "current") {
+    targetOpts.node = getCurrentNodeVersion();
+  }
+
+  if (targetOpts.hasOwnProperty("uglify") && !targetOpts.uglify) {
```
hzoo: If we didn't have to do this, would be good? doesn't seem necessary

README.md
```diff
@@ -122,6 +122,14 @@ A query to select browsers (ex: last 2 versions, > 5%) using [browserslist](http
 
 Note, browsers' results are overridden by explicit items from `targets`.
 
+### `targets.uglify`
+
+`number | true`
```
hzoo: although currently we only support uglify2

src/index.js
yavorsky: It will allow `uglify: null` and use it exactly like a `true`.  I think we must check other targets the same way.

README.md
yavorsky: @hzoo I think 1 also. No?

hzoo: I guess I just mean it's the same for whatever value

felixfbecker:
This doesn't seem to enable transpilation of `new.target` - I get an UglifyJS error
```
Unexpected token: punc (.)
```

hzoo:
@felixfbecker it wasn't supported at all in babel until recently, and we didn't add it to the preset yet

felixfbecker:
@hzoo Sorry, I assumed so because the PR was merged in July: https://github.com/babel/babel/pull/5906
What would I need to do to get it working?

hzoo:
It's actually because we never actually added it to the env/es2015 presets. I guess you could do it manually in `plugins` for now as a workaround

felixfbecker:
Shouldn't it be included through the class transform?

> The new.target property, is transpiled to this.constructor. As, with pure function constructor based prototype inheritance, the this instance is created by the child class, and given to the parent class constructor & methods (see super), this.constructor refers to the child class prototype constructor property, itself fixed to the child class function itself, matching to the constructor invoked by new.

https://www.npmjs.com/package/babel-plugin-transform-class#newtarget

Could I do a PR to add it?

PR #181, opened 2017-03-02
Drop whitelist option

yavorsky:


yavorsky:
@existentialism Anyway we could write some tests to make sure nothing happens if you pass `whitelist` 😜

PR #182, opened 2017-03-02
Fix running tests

existentialism:
Looks like we also need to either augment the custom test sandbox with the current node version, or change https://github.com/babel/babel-preset-env/blob/master/test/fixtures/preset-options/use-builtins-all-exec/exec.js#L1


existentialism:
Waiting on https://github.com/babel/babel/pull/5410

hzoo:
I'm ok with disabling the test for now to land the other prs?

PR #183, opened 2017-03-03
Prettify.

yavorsky:


danez:
Should we enable prettifier?

yavorsky:
@danez Yeah, I think we could prettify not separate fragments, but all the stuff. The only thing - we must be consistent with babel and babylon prettify branches.

xtuc:
Could we import the rules defined in https://github.com/babel/babel/pull/5413/files#diff-1dc6ee56b778cd91e0327b52aaeaa1b9? We will then include theses rules by default in babel-eslint-config.

yavorsky:
@xtuc Sure 👍 
Also disabled "arrow-parens" and "indent" to prevent errors after prettier.

package.json
```diff
@@ -81,5 +84,19 @@
...
+  "lint-staged": {
+    "src/*.js": [
+      "prettier --trailing-comma all --write \"src/*.js\"",
+      "git add"
+    ],
+    "scripts/*.js": [
+      "prettier --trailing-comma all --write \"scripts/*.js\"",
+      "git add"
+    ],
+    "test/*.js": [
```
existentialism: Probably want to exclude `expected.js`?

package.json
```diff
@@ -81,5 +84,19 @@
     "plugins": [
       "transform-flow-strip-types"
     ]
+  },
+  "lint-staged": {
+    "src/*.js": [
+      "prettier --trailing-comma all --write \"src/*.js\"",
```
existentialism: Can omit the `\"src/*.js\"` as --write will replace

existentialism: Also, can probably combine these to:

```json
{
  "lint-staged": {
    "{src,scripts}/*.js": [
      "prettier --trailing-comma all --write",
      "git add"
    ]
  }
}
```

package.json
yavorsky: @existentialism `test/*.js` will cover just files inside script dir, but not below test directory, so `fixtures`/`debug-fixtures` shouldn't be affected.

existentialism: :+1:

package.json
yavorsky: @existentialism thanks 👍

existentialism:
@yavorsky mind bumping to latest prettier and rerunning?

yavorsky:
@existentialism 👍

PR #184, opened 2017-03-03
Killer feature: Replace preset-es2015 with preset-env!

yavorsky:
One of these days after the second cup of coffee with milk and halva I discovered awesome preset that could be useful for us. I guess you haven't heard about it, but it's awesome! By specifying node version we could use only plugins we really need. Also, we could add builtIns if we want.

I hope you would like it!


existentialism:
giphy

PR #188, opened 2017-03-05
Clarify reasons for the uglify option in README.md

mikegreiling:
I found the `targets.uglify` option confusing when I first read the description until I dug into the issue.  I think this may help to clarify what it does and why it is necessary.


existentialism:
@mikegreiling thanks for the PR! Can we also add a recommendation for Babili (https://github.com/babel/babel-preset-env/issues/190)?

mikegreiling:
@existentialism I've updated my PR to include a link to Babili

existentialism:
@mikegreiling thanks again!

PR #191, opened 2017-03-06
Target electron current

mozmorris:
This is a first attempt at trying to support `electron: "current"` as a target. Hopefully this is something others might be interested in. 😬

It works in a similar way to `node: "current"` and targets the locally installed Electron version.

Somethings to note or discuss:

1. I've added Electron as a dev dependency but ideally this wouldn't be included at all. Perhaps I could stub the output from `npm list electron` for testing.

2. I'm using `execSync` to run `npm list electron` which slows down the tests a fair bit. As above, maybe this could be stubbed.

3. Is running `npm list electron` the best way to get the locally installed version of Electron. Maybe it would be better to read this from `package.json`


hzoo:
Yeah adding electron as a devDep doesn't seem necessary

cc @paulcbetts, @Kilian

test/index.spec.js
```diff
@@ -2,7 +2,7 @@
 
 const babelPresetEnv = require("../lib/index.js");
 const assert = require("assert");
-const { versions: electronToChromiumData } = require("electron-to-chromium");
+const electronToChromiumData = require("electron-to-chromium");
```
Kilian: Perhaps you can do

`const electronToChromiumVersions = require("electron-to-chromium/full-versions");`

Since you're not using other data (This keeps it a little smaller)

test/index.spec.js
```diff
@@ -83,6 +83,14 @@ describe("babel-preset-env", () => {
         }, Error);
       });
     });
+
+    it("should return the current locally installed Electron version with option 'current'", function() {
+      assert.deepEqual(babelPresetEnv.getTargets({
+        electron: "current"
+      }), {
+        chrome: parseInt(electronToChromiumData.electronToChromium("1.6.1"), 10)
```
Kilian: With requiring just the list of full electron versions, `electronToChromiumVersions["1.6.1"]` would be then be the way to get to this info.

paulcbetts:
> Is running npm list electron the best way to get the locally installed version of Electron. Maybe it would be better to read this from package.json

I would try to get the path of the top-level `package.json`, then get `node_modules/electron/package.json` relative to that

mozmorris:
@Kilian have made those small changes.

@paulcbetts Yeah I think for speed that's probably a better way to go.

I'm now returning a hardcoded version number during test. This means we don't need to add Electron as a dependency. I'm open to suggestions if you think this could be done in a better way.

mozmorris:
On reflection I don't think using `process.env.NODE_ENV` is a good idea.

hzoo:
Yeah I think running execSync with npm is pretty slow

src/index.js
```diff
@@ -80,6 +81,17 @@ export const getCurrentNodeVersion = () => {
   return parseFloat(process.versions.node);
 };
 
+/**
+ * Returns the locally installed Electron version
+ * @return {string}
+ */
+export const getLocalElectronVersion = () => {
+  // Electron is not available in test environment so mock the version
+  if (process.env.NODE_ENV === "test") return "1.6.2";
```
hzoo: Can this just be mocked in the tests, I don't think it needs to be in the code. ^ I don't think you are using the method otherwise even though you made it an export?

mozmorris: Yeah I'm going to remove that but the problem still exists, how can I mock the result of `getLocalElectronVersion` in test?

I'm calling it in `getTargets`. An option would be to inject it into `getTargets`?.

mozmorris:
Now using package.json of locally installed Electron to get version number instead of `npm list electron`. This means I can mock the package.json in test and test that an error is thrown when Electron isn't present.

I haven't squashed these commits yet but let me know if you want me to.

test/index.spec.js
```diff
@@ -22,67 +24,103 @@ describe("babel-preset-env", () => {
...
+        if (fs.existsSync(electronDir)) fs.removeSync(electronDir);
+        fs.mkdirSync(electronDir);
+        fs.copySync(source, destination);
+      });
+
+      after(() => {
+        fs.remove(electronDir);
+      });
+
+      it("should return the locally installed Electron version", function() {
```
existentialism: Mind also adding a test for when local electron version could not be found?

mozmorris: That diff isn't too clear, but I'm also testing for that [here](https://github.com/babel/babel-preset-env/pull/191/files#diff-faa503f61f8addeb603095cc2f40c82eR91)

existentialism: Yep!

mozmorris: I've made those installed/not-installed tests a little clearer.

mozmorris:
Updated so that `"electron": true` also target local version. Docs updated.

yavorsky:
@mozmorris 
Thanks for the PR! Sorry for the late review.

Just wondered, if there a way to get global electron version?
I'm not very familiar with the electron API, but, as I understand, in node.js environment `require('electron')` will return electron's path. Having it we could get the `package.json` of globally installed electron location and read the version. Please correct me if I wrong.

It could be incomprehensible, if user has only global electron's version, use it, specify `"current"` version and get an error `A local version of Electron was not found`.
With this way, if I'm wrong and it's really impossible to read global version, would be nice to extend error message and notify people to add electron as dependency in `package.json`.

mozmorris:
@yavorsky that seems fair that we should fallback to the globally installed version if available. I think it's uncommon but we should still handle that use case.

I'm just not sure how I could write a test for this without adding Electron as a dependency of babel-preset-env.

yavorsky:
@mozmorris The advantage of global dependency check is that we don't need to add it as dependency in package.json 😊
You've wrapped `getLocalElectronVersion` body in try/catch, so adding just `require('electron')` in try block has the similar susceptibility to errors like current `require(${rootPath.path}/node_modules/electron/package.json)`.

mozmorris:
Sorry I wasn't very clear. I can definitely add the functionality but was concerned about testing it.

I'm not sure we can write a test for that globally installed use case?

yavorsky:
@mozmorris Oh, got it. 🙂
Seems like we just could 1) mock require('electron') and return path like it does in node.js environment or 2) add it as devDep.
Electron size 120+mb and it could be really annoying to update devDependencies. I'm not sure, but maybe let's start with the option 1.

mozmorris:
Yeah, I added electron as a dev dep on the initial PR, that sucked! 😉

I'll look at mocking require('electron'). That would probably also work for the locally installed version too.

mozmorris:
Thanks for the review btw 👍

yavorsky:
@mozmorris Thanks for PR! 👍👍👍
Going to try to help with updates.

mozmorris:
Just had a chance to take a look at this.

```
> require('electron')
'/Users/Moz/Code/aproject/node_modules/electron/dist/Electron.app/Contents/MacOS/Electron'
```

This is the path to the Electron executable, which is different across OS's. The package.json in this example is 5 directories above.

Then there's also `require.resolve('electron')`

```
> require.resolve('electron')
'/Users/Moz/Code/aproject/node_modules/electron/index.js'
```

This is the directory with the package.json. However I don't know how I could mock `require.resolve('electron')` in test.

existentialism:
Not _super_ elegant, but could move `resolve.require("electron")` call to a diff file and use something like proxyquire to replace it w/ a stub that returns a path to local mock test file?

test/index.spec.js
Asudhaishao: 需要好好学习了，特么的兴奋看到你是怎么做的？

yavorsky: @Asudhaishao Хто його зна..

hzoo:
Thanks for the PR! 

PR #192, opened 2017-03-06
Add transform-duplicate-keys mapping

existentialism:
Thanks to @Timer, it looks like we missed mapping this transform causing output to be different than `preset-latest`.

I double checked that the rest of the plugins are mapped.

Ref: https://github.com/facebookincubator/create-react-app/pull/1742


Timer:
You're a saint. Thanks!

yavorsky:
@hzoo We are not testing all plugins (just opts for `regenerator` and `transform-polyfill-require-plugin`), so I think it's not necessarily. Updating `stdout.txt` should be ok. Or we could add tests to cover all or most of plugins.
Mabye let's test matching latest without targets?

Timer:
Anything I can do to help?

existentialism:
Added a quick and dirty sanity check test, it doesn't hurt to have these until we move to more comprehensive tests on Browserstack or some other means.

PR #195, opened 2017-03-08
Add built-ins, better links, compat-table url, etc

yavorsky:
1. Fix label for plugin-features.
2. compat-table URLs target directly to the feature.
3. Typo fixes.
4. Add built-in-features section.
5. mention that we are using compat-table with the latest commit hash.


hzoo:
Awesome 👍

PR #200, opened 2017-03-10
Enable code coverage

alxpy:
I propose to add code coverage to this repo.
Can someone from members enable integration with `codecov.io`?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 node_modules
+coverage
 lib
 test/tmp
 .DS_Store
 *.log
 .vscode
+.nyc_output
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,11 @@ node_js:
 - '0.12'
 - '0.10'
 env:
-  - PKG_CMD="npm"
-  - PKG_CMD="yarn"
+  global:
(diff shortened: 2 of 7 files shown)


alxpy:
> codecov/project — No report found to compare against

I think it's OK for the first PR with `codecov`

existentialism:
Couple things:

- Probably want to _only_ include `src`, no sense in doing coverage on the data files, etc.

- Should set `all: true` to include any files that don't have any tests

- To get better coverage results, let's instrument the code via `babel-plugin-instanbul` (via BABEL_ENV/NODE_ENV).

alxpy:
@existentialism done.
But it looks like `babel-plugin-instanbul` does not work with `Node.js 0.10/0.12`. Should we support the `Node.js 0.10/0.12`?

existentialism:
@alxpy for master we should use 3.1.2, and can bump to 4.0 on `2.0`, if you don't mind making that change, I'll follow it up with some tweaks for CI/documentation.

existentialism:
@alxpy made a few tweaks, please review!

alxpy:
@existentialism LGTM!

existentialism:
@alxpy thanks again!

PR #201, opened 2017-03-10
Bump plugins

yavorsky:
Ok, I'm cooking tests to compare preset-latest vs preset-env without options. It would be useful while we suggesting to switch to the latest and to prevent bugs like https://github.com/babel/babel-preset-env/pull/192, and for proper comparison, we need latest versions.

Also, rebuild replaced duplicate-keys according to alphabetical order.


existentialism:
@yavorsky mind rebasing?

yavorsky:
@existentialism should be ok 😊

PR #86, opened 2016-12-12
use interpolate function from compat-table fixes #45

graingert:

PR #208, opened 2017-03-14
Refactor browser data parsing to handle families

### LINKED ISSUE #207: Edge 14 doesn't get block scoping detected
Edge 14 passes all tests in the Kangax ES6 table in the `const` & `let` sections:
http://kangax.github.io/compat-table/es6/

However, specifying `Edge >= 14` as a target still transpiles `const` to `var`:
https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=env&targets=&browsers=Edge%20%3E%3D%2014&builtIns=false&experimental=false&loose=false&spec=false&playground=false&code=const%20x%20%3D%202%3B%0A

I've looked at the code base. This is what I found:

1. Edge is not included in the list of environments that don't require the `es2015-block-scoping` transform: https://github.com/babel/babel-preset-env/blob/v1.2.1/data/plugins.json#L20-L27
2. The `es2015-block-scoping` transform seems to rely only on the results of `const` & `let` tests if I undersdand the code correctly: https://github.com/babel/babel-preset-env/blob/v1.2.1/data/plugin-features.js#L13-L18

It seems to me all conditions for Edge 14 to be accepted should be met?

existentialism:
Fixes #207.

And by refactor, it just means I just adapted compat-table's code for interpolating environments (essentially what #86 was doing).

One nice side effect is that this allows us to drop the `invertedEqualsEnv` stuff, and may allow for some additional cleanup/hacks but will wait on that.

/cc: @mgol hope you weren't too far towards a fix, I was just looking at some code and realized it was a quick fix, would appreciate your eyes on this for a review!


mgol:
I haven't started preparing the actual PR yet, fortunately. :) I'll have a look at the PR.

scripts/build-data.js
```diff
@@ -6,16 +6,88 @@ const path = require("path");
...
+const interpolateAllResults = (rawBrowsers, tests) => {
+  const interpolateResults = (res) => {
+    let browser;
+    let prevBrowser;
+    let result;
+    let prevResult;
+    let bid;
+    let prevBid;
+
+    for (bid in rawBrowsers) {
```
mgol: This could be `for (const bid in rawBrowsers)` and then the above `let bid` declaration is not needed.

scripts/build-data.js
```diff
@@ -118,31 +161,26 @@ const getLowestImplementedVersion = ({ features }, env) => {
 
   const envTests = tests
     .map(({ res: test, name, isBuiltIn }, i) => {
+
+      if (name === "const/basic support") {
```
mgol: A debugging leftover?

mgol:
Overall looks good, I have some minor remarks.

scripts/build-data.js
```diff
@@ -118,31 +161,26 @@ const getLowestImplementedVersion = ({ features }, env) => {
...
-      .filter((t) => t.startsWith(env))
-      // Babel assumes strict mode
-      .filter((test) => tests[i].res[test] === true || tests[i].res[test] === "strict")
-      // normalize some keys
-      .map((test) => envMap[test] || test)
-      .filter((test) => !isNaN(parseFloat(test.replace(env, ""))))
-      .shift();
+        .filter((t) => t.startsWith(env))
+        // Babel assumes strict mode
+        .filter((test) => tests[i].res[test] === true || tests[i].res[test] === "strict")
```
yavorsky: We could merge 2 filters in 1 for perf, but it's just a nit and could be changed in separate PR aimed to optimize things like this.

scripts/build-data.js
existentialism: Yep lol

PR #209, opened 2017-03-14
Test for latest matching.

yavorsky:
Add test case to compare all latest plugins with all env plugins.
As we know, env preset without options must has the same behavior as a latest.
I think it might be useful while we recommend change latest to env.
Will prevent cases like https://github.com/babel/babel-preset-env/pull/192.


hzoo:
Thanks for the PR, although given we are deprecating preset-latest I didn't think we would be updating it anymore?

yavorsky:
@hzoo Yes! I consider this PR as a temporal solution for the transition period.

PR #211, opened 2017-03-14
Fix transform-polyfill-require transformed paths

romulof:
It was importing stuff from outside `library` folder.

### DIFF
--- a/src/transform-polyfill-require-plugin.js
+++ b/src/transform-polyfill-require-plugin.js
@@ -32,7 +32,7 @@ export default function ({ types: t }) {
 
   function createImport(polyfill, requireType, core) {
     if (core) {
-      polyfill = `core-js/modules/${polyfill}`;
+      polyfill = `core-js/library/modules/${polyfill}`;
     }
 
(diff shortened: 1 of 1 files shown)


yavorsky:
Hi, @romulof, thanks for the PR!

Trying to understand why it's better to use 'library/modules' location. Did you have some troubles with imports?

If yes, we must fix all the tests which requires 'core-js/modules/[feature]' imports.

romulof:
All other libraries and dependencies import from `core-js/library/...`, so when preset-env imports from `core-js/modules` some modules get duplicated.

The original core-js documentation states to import from `library` dir.

This simple fix has shrunk my bundle ~10kB.

hzoo:
cc @zloirock so we should be using /library? I didn't know

zloirock:
Nope, here should be `core-js/modules`, not `core-js/library/modules`. `library` is a version without global namespace pollution, but it's a regular polyfill.

romulof:
@zloirock, thanks for the explanation. I'll see which libraries are using the "wrong" version.

PR #223, opened 2017-03-23
remove deprecated projects

stevemao:


hzoo:
Although I kinda just asked everyone to deprecate in favor

PR #225, opened 2017-03-24
Run yarn as pre-commit hook if package.json was changed

yavorsky:
Looks like not only me is tired of adding yarn.lock every time when some dependencies were changed. I was thinking about it and have an idea to run `yarn` as a pre-commit hook like we are doing for prettier. To run it only when `package.json` was added to commit. (thanks for awesome feature by @okonet).

What do you think?

### DIFF
--- a/scripts/yarn-install.js
+++ b/scripts/yarn-install.js
@@ -1,6 +1,6 @@
 "use strict";
 
-const { exec } = require("child_process");
+const exec = require("child_process").exec;
 
 const runIfYarn = fn => {
   exec("yarn -V", error => {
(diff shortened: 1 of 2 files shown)


scripts/yarn-install.js
```diff
@@ -0,0 +1,13 @@
+"use strict";
+
+const { exec } = require("child_process");
```
okonet: Destructuring won't work on Node 4.x AFAIK. Also consider using https://github.com/sindresorhus/execa instead for better cross OS support.

yavorsky: @okonet accidentally pushed the twitter version 😅

Issue #220, opened 2017-03-20
exclude 'web' polyfills from node by default


restrry:
Hi, after running `babel-env-preset`  with `useBuiltIns: true` for node (v7.6) environment got in console:
```
Using targets:
{
  "node": 7.6
}

Modules transform: false

Using plugins:
  syntax-trailing-function-commas {"node":7.6}

Using polyfills:
  es7.string.pad-start {"node":7.6}
  es7.string.pad-end {"node":7.6}
  web.timers {"node":7.6}
  web.immediate {"node":7.6}
  web.dom.iterable {"node":7.6}
```
So, I was surprised to find 'web' polyfills included in the final node build (I use webpack for both node and web). Of course I could use `exclude` option, but probably it should be default behaviour for `node` env.

hzoo:
Sure we can do that. Do you want to make a PR?

existentialism:
Fixed by #226

PR #226, opened 2017-03-24
separate default builtins for platforms

restrry:
issue https://github.com/babel/babel-preset-env/issues/220

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,6 @@
 import browserslist from "browserslist";
 import builtInsList from "../data/built-ins.json";
-import defaultInclude from "./default-includes";
+import { defaultWebIncludes } from "./default-includes";
 import moduleTransformations from "./module-transformations";
 import normalizeOptions, { getElectronChromeVersion } from "./normalize-options.js";
 import pluginList from "../data/plugins.json";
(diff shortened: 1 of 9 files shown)


src/index.js
```diff
@@ -161,6 +161,14 @@ export const transformIncludesAndExcludes = (opts) => ({
   builtIns: opts.filter((opt) => opt.match(/^(es\d+|web)\./))
 });
 
+function getPlatformSpecificDefaultFor(targets) {
+  const targetNames = Object.keys(targets);
+  const isAnyTarget = !targetNames.length;
+  const isWebTarget = targetNames.some((name) => name !== "node");
+
+  return (isAnyTarget || isWebTarget) ? defaultInclude : [];
```
restrry: should we rename it?

existentialism: Maybe `defaultInclude` -> `defaultWebIncludes`?

hzoo: yeah since `import defaultInclude from "./default-includes";`

src/index.js
```diff
@@ -161,6 +161,14 @@ export const transformIncludesAndExcludes = (opts) => ({
   builtIns: opts.filter((opt) => opt.match(/^(es\d+|web)\./))
 });
 
+function getPlatformSpecificDefaultFor(targets) {
+  const targetNames = Object.keys(targets);
+  const isAnyTarget = !targetNames.length;
+  const isWebTarget = targetNames.some((name) => name !== "node");
```
yavorsky: maybe check for `electron` also?

hzoo: but you would want to include it for electron/uglify right?

yavorsky: If we have node + electron or node + uglify, isWebTarget will be true. But we don't need to include it in these cases ?

PR #229, opened 2017-03-28
Fill data with electron as a target.

yavorsky:
For now, we are converting electron version to chrome with `electronToChromium` when building the plugins list because of `compat-table` has no info about `electron`.

With this PR we could convert all chrome version we need to support into electron like we always do for the `opera`.

It will make consistent log output, better `preset-env` debugging, performance optimization (less work while building plugins for the preset). Also `data/plugins` or `data/built-ins` is a solid universal source that could be useful for different things in the future, so filling it is a good idea as for me.


scripts/build-data.js
```diff
@@ -7,9 +7,36 @@ const flatten = require("lodash/flatten");
...
+    return all;
+  }
+, {});
+const chromiumToElectronVersions = Object.keys(chromiumToElectronMap);
+
+const findClosestElectronVersion = (targetVersion) => {
+  const chromiumVersionsLength = chromiumToElectronVersions.length;
+  const maxChromium = +chromiumToElectronVersions[chromiumVersionsLength - 1];
+  if (targetVersion > maxChromium) return null;
+  const closestChrome = chromiumToElectronVersions.reduce((result, version) => {
```
existentialism: This results an incorrect version of electron being mapped, for example for async:

```json
{
  "chrome": 56,
  "electron": 1.5
}
```

Electron should be `1.6`.

Should be able to reduce (hehe) the reduce with something like:

```js
const closestChrome = chromiumToElectronVersions.find((version) => targetVersion <= version);
```

**EDIT:** even better, I guess we can drop `findClosestElectronVersion` completely and use `chromiumToElectron` [once it lands](https://github.com/Kilian/electron-to-chromium/pull/8)?

yavorsky: @existentialism so we need to update `chromiumToElectron`. for now, it will just return `undefined` if no explicit values were found.

Kilian:
1.3.0 is released! Thanks @yavorsky :) https://www.npmjs.com/package/electron-to-chromium

hzoo:
@Kilian there's a regression with 1.3.0 ` Cannot find module './full-chrome-versions'`

Kilian:
v1.3.1 up now, sorry about that.

yavorsky:
@Kilian please, join babeljs.slack.com

malept:
FYI, this broke [Electron Forge](https://github.com/electron-userland/electron-forge), because we assume that we can send a full Electron version (as a string) as a target version. Downgrading from 1.3.0 to 1.2.2 fixes it. I need to see if https://github.com/babel/babel-preset-env/pull/231 fixes this for us.

yavorsky:
@malept Can you use float numeric value? `1.2 instead of "1.2.0" for ex.?`

Also, https://github.com/babel/babel-preset-env/pull/231 will allow strings as version values.
I think we are going to merge it in the near future.

MarshallOfSound:
@yavorsky Might have been inconsistent but this change should have been considered breaking and been bumped as such IMO 👍

yavorsky:
@MarshallOfSound It was rather bug fix than breaking change 🙂. [According to the docs](https://github.com/babel/babel-preset-env#options ), target must be only a number (unless we merge https://github.com/babel/babel-preset-env/pull/231).

But yeah we should add this at least to changelog.

jquense:
broke me too. I realize that the "correct" handling of this is a bit vague, but generally it'd be good side on the one that doesn't break folks, even if it was technically bug. Not a huge deal but just adding my voice. Thanks ya'll

existentialism:
@MarshallOfSound @jquense sorry for the unintentional breakage, we restored string version support for electron in https://github.com/babel/babel-preset-env/releases/tag/v1.3.3!

Issue #187, opened 2017-03-03
Version numbers should be string not numbers


danez:
As we are currently using numbers for targets, it can happen that too much plugins get applied although not necessary.

```js
{
  node: 6.10,
}
```

This will apply unnecessary transforms as 6.1 < 6.9

```
Using targets:
{
  "node": 6.1
}

Modules transform: commonjs

Using plugins:
  transform-es2015-destructuring {"node":6.1}
  transform-es2015-for-of {"node":6.1}
  transform-es2015-function-name {"node":6.1}
  transform-exponentiation-operator {"node":6.1}
  transform-async-to-generator {"node":6.1}
  syntax-trailing-function-commas {"node":6.1}
```

danez:
#114

spencerhakim:
The [new `invertedEqualsEnv` code](https://github.com/babel/babel-preset-env/pull/180/files#diff-b7a8695388d01f5a44747b19feee2e3dR54) from #180 should be easy to modify to instead use the `semver` package, no? So that we don't have to wait for 2.0 or for #114 to be fixed and merged. I can submit a PR if that'd help.

danez:
@spencerhakim Yes true, but I'm not talking about `invertedEqualsEnv` but the user supplied options. But it will be fixed in #114.

But good catch we also need to use semver in `invertedEqualsEnv` though it cannot fail there currently as compat-table has no version '>=x.10' defined.

yvele:
I juste made https://github.com/babel/babel-preset-env/issues/227 but I'm closing it as it seems to be a duplicate of this one..

damonmaria:
I suspect this issue will become a lot more common now that AWS Lambda is running Node 6.10.

existentialism:
We landed string support in https://github.com/babel/babel-preset-env/issues/231.

I'll open another issue for supporting things like `"chrome": ">= 53"` separately!

PR #231, opened 2017-03-28
Support target versions as strings

existentialism:
First part of #187.

Wanted to keep it relatively small, so only the following changes:

- Added `semverify` and moved `_extend` to utils file ala https://github.com/babel/babel-preset-env/pull/114
- `build-data` now outputs string values for env versions
- String target versions are now allowed via using `semver` to check if a plugin is required
- Left backwards compatibility (though cases like plugin `"node": 6.9` and target `"node": 6.10` will behave differently) with numeric target versions, not sure if we should drop in 2.0?

We should follow this up with https://github.com/babel/babel-preset-env/pull/114 then possibly add range support to targets (`"chrome": "^52"`) if we want.


yavorsky:
Maybe semverify all targets' versions immediately? If there are no difference between '5' and 5, why we should log it as a different values? Also it would be good for the `flow` brunch to have the same version type during the whole preset process. 😏

existentialism:
"Few" more changes:

- Normalizing (into semver) version now occur at `getTargets`, so they're strings all the way down.
- Moved `getTargets` into `targets-parser`, and refactored it a bit.
- `isPluginRequired` now assumes passed environments have already been parsed by `getTargets`, so it doesn't do checks or run `getTargets` itself
- For cleaner debugging output, added some utils to drop extraneous `.0` from versions

scripts/build-data.js
```diff
@@ -245,20 +245,22 @@ const generateData = (environments, features) => {
     environments.forEach(env => {
       const version = getLowestImplementedVersion(options, env);
       if (version !== null) {
-        plugin[env] = version;
+        plugin[env] = version.toString();
```
existentialism: Should we `semverify` these?

Issue #84, opened 2016-12-12
useBuiltIns: figure out how to remove polyfills that aren't actually used in the codebase


hzoo:
### Don't include if supported environments support it natively #20
- Use preset-env config to determine what parts of core-js to pull in (original PR, #55)

```js
import "babel-polyfill";
```

gets transformed to

```js
import "core-js/modules/es6.map";
import "regenerator-runtime/runtime";
...
```

### Don't include if you have blacklisted it yourself (regenerator for example) #60
- Be able to blacklist transforms or specific builtIns? 

```js
{
  "presets": [
    ["env", {
      "targets": {
        "browsers": ["last 2 versions", "safari >= 7"]
      },
      "blacklist": [
        "transform-regenerator" // won't include "regenerator-runtime/runtime"
      ]
    }]
  ]
}
```

---

### Don't include if the codebase doesn't actually use the builtIn (this)

> this would need to be a whole codebase transformation so it probably won't work in this preset/plugin

The problem with the first heading ^ is that even if you remove the polyfills supported natively in the environment, you might not actually use them in your codebase anyway.

Possible solution: run an analyzer (separate babel pass like [transform-runtime](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-runtime)) on the whole codebase and remove imports that aren't used anywhere.

EDIT: A big problem atm is that Babel runs per file. So unless we understand the whole codebase it sounds like something webpack could do? cc @TheLarkInn @keyanzhang

TheLarkInn:
So here is an idea (atleast in terms of webpack). //cc @patrickkettner of Modernizr who this could be highly relevant for.

## On demand async polyfill chunks

### Background: ContextModules with webpack 

Since webpack is a static build tool, there is no _real_ dynamic requires, etc. However, there is a static solution to this dynamic nature. 

Given the following builtins are statically located somewhere:

```
--/builtins
  --/somePoly1.js
  --/somePoly2.js
  --/somePoly3.js
```

One can pass an expression to a dependency statement. Specifically `System.import()` or `import()` (for async chunking). 

```javascript 
function asyncLoadPoly(builtInUniqueName) {
  return `import('./builtins' + builtInUniqueName)`
}
```

webpack will create a separate chunk for each possible module that could be loaded from that folder.

This will then return a Promise and one can load that specific poly at runtime: 

```javascript
if (someLogicAboutSpecificJSFeature) {
  asyncLoadPoly('somePoly1.js').then(
[...]

PR #241, opened 2017-03-31
Breaking: make current "useBuiltIns" auto import only used + necessary polyfills per file

hzoo:
https://twitter.com/left_pad/status/847911365645938690

Ref https://github.com/babel/babel-preset-env/issues/84

TL;DR: automatically **add** polyfills based on actual usage in the file (check the presence of `Promise`), and **remove**  any polyfills supported by the target environment (Chrome 55 already supports `Promise`). Check the tests in this pr for more examples!

Size savings: if you don't use any polyfills in your code: then none will be added. If you do use polyfills, only the ones not supported in your target environment will be added.

## Older browser (like IE): imports are added

screen shot 2017-03-31 at 4 12 39 pm

## Modern browser (Chrome 58): no imports added

screen shot 2017-03-31 at 4 17 53 pm

---

astexplorer live example (only the adding part): http://astexplorer.net/#/gist/dc031f702921a727ec5f82c2bc3341b0/6b87729fb2ed77e8512c0b137e8a3c19dbcdc63c

---

The only issue not covered in https://github.com/babel/babel-preset-env/issues/84#issuecomment-267235999 is regarding the rest of node_modules (3rd party).

Basically the issue is that if you don't use `Promise` in any of the code that you run Babel on, this option doesn't know to include the `Promise` polyfill if a library you use needs it.

A workaround is a disclaimer of the option and to use `includes` instead. We can even warn if you include but it's natively supported.

---

Not sure about the option name, or if we should making a breaking change for it. It does seem like this should be opt-in since it doesn't know about the use of polyfills in 3rd party modules unless you run babel over those too

- [x] more tests (suggest some more?)
- [x] change the console.warn on instance methods to be part of the debug flag?
- [x] change option to be useBuiltIns, and move current option to be `useBuiltIns: "entry"`
- [x] fix import not being transform (due to use of Program.exit, should change to adding the imports in the visitors themselves)
- [x] Can we track VariableDeclaration with StringLiteral matched to built-in definition and in case of a[trackedVariable] include built-in according to StringLiteral value?


test/fixtures/preset-options-add-used-built-ins/test-modules-tranform/expected.js
```diff
@@ -0,0 +1,4 @@
+"use strict";
+
+import "core-js/modules/es6.promise";
```
hzoo: bug: not transformed due to program.exit

src/built-in-definitions.js
```diff
@@ -0,0 +1,108 @@
+export const definitions = {
```
hzoo: This is basically the opposite of built-in-features.js so maybe can just reverse it somehow

src/index.js
```diff
@@ -212,8 +213,11 @@ export default function buildPreset(context, opts = {}) {
     [require(`babel-plugin-${pluginName}`), { loose }]
   ));
 
-  useBuiltIns &&
+  if (useBuiltIns) {
     plugins.push([transformPolyfillRequirePlugin, { polyfills, regenerator }]);
+  } else if (addUsedBuiltIns) {
+    plugins.push([addUsedBuiltInsPlugin, { polyfills, regenerator }]);
```
hzoo: right now it's a separate option/plugin/file to make it easier to look at

yavorsky:
Can we track `VariableDeclaration` with `StringLiteral` matched to built-in definition and in case of `a[trackedVariable]` include built-in according to `StringLiteral` value?

hzoo:
Yeah we can just dono if it's worth doing it

hzoo:
new idea:

Make this the default behavior in 2.0 (useBuiltIns: true) which is "aggressive" in removing polyfills that aren't used in the files. Another option would be the current `useBuiltIns` which only removes polyfills already available in the environment. And lastly `useBuiltIns: false` which doesn't do anything.

hzoo:
9f754cf just appends `babel-polyfill/` so it requires the same module (otherwise it requires you to have core-js in your dependencies when likely you have babel-polyfill or the wrong version of core-js)

README.md
```diff
@@ -176,41 +174,71 @@ This option is useful for "blacklisting" a transform like `transform-regenerator
...
-This will also work for `core-js` directly (`import "core-js";`)
+```js
+var b = new Map();
+```
+
+#### `useBuiltIns: 'entry'`
+
+> NOTE: Only use `require("babel-polyfill");` once in your whole app. One option is to create a single entry file that only contains the require statement.
+
+This option enables a new plugin that replaces the statement `import "babel-polyfill"` or `require("babel-polyfill")` with individual requires for `babel-polyfill` based on environment.
```
existentialism: Worth adding a note this was the behavior in 1.x?

src/add-used-built-ins-plugin.js
```diff
@@ -0,0 +1,245 @@
...
+  const addAndRemovePolyfillImports = {
+    ImportDeclaration(path, state) {
+      if (
+        path.node.specifiers.length === 0 &&
+        isPolyfillSource(path.node.source.value)
+      ) {
+        state.opts.addUsedBuiltIns &&
+          console.warn(
+            `
+Adding "import 'babel-polyfill'" isn't necessary with the addUsedBuiltIns option anymore.
```
existentialism: `addUsedBuiltIns ` -> `useBuiltIns: true`

README.md
existentialism: Eh, maybe not and just put it in upgrade guide.

src/normalize-options.js
```diff
@@ -76,6 +76,6 @@ export default function normalizeOptions(opts) {
     loose: validateLooseOption(opts.loose),
     moduleType: validateModulesOption(opts.modules),
     targets: opts.targets,
-    useBuiltIns: opts.useBuiltIns,
+    useBuiltIns: opts.useBuiltIns === undefined ? true : opts.useBuiltIns,
```
hzoo: breaking: default to true!

src/use-built-ins-entry-plugin.js
```diff
@@ -1,5 +1,5 @@
 function isPolyfillSource(value) {
-  return value === "babel-polyfill" || value === "core-js";
```
hzoo: since the 7.0 babel-polyfill will have core-js

src/use-built-ins-plugin.js
```diff
@@ -0,0 +1,227 @@
...
+
+  const addAndRemovePolyfillImports = {
+    ImportDeclaration(path) {
+      if (
+        path.node.specifiers.length === 0 &&
+        isPolyfillSource(path.node.source.value)
+      ) {
+        console.warn(
+          `
+Adding "import 'babel-polyfill'" isn't necessary with the useBuiltIns option anymore.
```
existentialism: Maybe:

> Adding "import 'babel-polyfill'" is no longer necessary when setting `useBuiltIns: true` since polyfills are automatically added when needed. Please remove the call or use `useBuiltIns: 'entry'` instead.

?

existentialism:
Feel like we may want to move debug output logic to the transform is being used. Currently, "Using plugins" and "Using polyfills" with the new `useBuiltIns: true` is a bit misleading since they may or may not actually have been included based on usage.

README.md
```diff
@@ -176,41 +174,71 @@ This option is useful for "blacklisting" a transform like `transform-regenerator
...
-> This will only work with npm >= 3 (which should be used with Babel 6 anyway)
+A way to apply `babel-preset-env` for polyfills (via `babel-polyfill`).
 
-```
+```sh
 npm install babel-polyfill --save
 ```
 
-This option enables a new plugin that replaces the statement `import "babel-polyfill"` or `require("babel-polyfill")` with individual requires for `babel-polyfill` based on environment.
+#### `useBuiltIns: true`
```
mikesherov: Are you missing docs for useBuiltIns: false now? May seem obvious, but now that it's not the default you should at least describe the option.

src/index.js
```diff
@@ -242,8 +242,11 @@ export default function buildPreset(context, opts = {}) {
     ]),
   );
 
-  useBuiltIns &&
-    plugins.push([transformPolyfillRequirePlugin, { polyfills, regenerator }]);
+  if (useBuiltIns === true) {
+    plugins.push([addUsedBuiltInsPlugin, { polyfills, regenerator, debug }]);
```
mikesherov: stray debug?

src/normalize-options.js
mikesherov: Maybe convert this to a destructure with a default?

mikesherov: Along with opts.target and opts.debug

src/use-built-ins-plugin.js
```diff
@@ -0,0 +1,227 @@
...
+            addUnsupported(path, state.opts.polyfills, builtIn, this.builtIns);
+          }
+        }
+
+        if (
+          !node.computed &&
+          t.isIdentifier(prop) &&
+          has(definitions.instanceMethods, prop.name)
+        ) {
+          state.opts.debug && warnOnInstanceMethod(getObjectString(node));
```
mikesherov: Maybe move debug check into warnOnInstanceMethod?

src/use-built-ins-plugin.js
```diff
@@ -0,0 +1,227 @@
...
+      importDec._blockHoist = 3;
+      const programPath = path.find(path => path.isProgram());
+      programPath.unshiftContainer("body", importDec);
+    }
+  }
+
+  function addUnsupported(path, polyfills, builtIn, builtIns) {
+    if (Array.isArray(builtIn)) {
+      for (const i of builtIn) {
+        if (polyfills.indexOf(i) !== -1) {
```
mikesherov: Maybe polyfills could be a `Set` so you get O(1) lookups instead of O(n)? Might be premature optimization.

src/use-built-ins-plugin.js
```diff
@@ -0,0 +1,227 @@
...
+the "polyfills" option was not correctly passed
+to the "transform-polyfill-require" plugin
+`,
+          );
+        }
+        path.get("body").forEach(bodyPath => {
+          if (isRequire(bodyPath)) {
+            console.warn(
+              `
+Adding "require('babel-polyfill')" isn't necessary with the useBuiltIns option anymore.
```
mikesherov: Avoid "anymore". This message is valid for new and old users. To users it *never* was necessary

src/use-built-ins-plugin.js
```diff
@@ -0,0 +1,227 @@
...
+        if (
+          !node.computed &&
+          t.isIdentifier(prop) &&
+          has(definitions.instanceMethods, prop.name)
+        ) {
+          state.opts.debug && warnOnInstanceMethod(getObjectString(node));
+          const builtIn = definitions.instanceMethods[prop.name];
+          addUnsupported(path, state.opts.polyfills, builtIn, this.builtIns);
+        } else if (
+          node.computed &&
```
mikesherov: This is very cool that it catches: `b['find']` but it won't catch `var c = 'find'; b[c];`. 

And I know it's literally impossible to statically guarantee that you'll find all cases because of stuff like `var c = 'fi' + 'nd'; b[c];`

However, maybe it makes sense to iterate over all string literals and if any match an instance method name, include the polyfill?

I only say that because it's far more likely to see: `b[someCondition ? 'find' : 'findIndex']` and you never see the literal string form because you can use non computed *always* in that case.

src/index.js
existentialism: The debug flag is used in the plugin => https://github.com/babel/babel-preset-env/blob/ebf77d82181ac0be44153cd2041a9cdd9a0c4e08/src/use-built-ins-plugin.js#L142

src/normalize-options.js
existentialism: We may also want to validate valid `useBuiltIn` values (`false|true|'entry'`)

src/index.js
hzoo: Maybe we don't want to have those warnings, not sure if it's useful or not but we can iterate

src/use-built-ins-plugin.js
hzoo: It's passed in as an array, so can turn it into a set before passing it down

existentialism: Might be better to review using a `Set` throughout the plugin in 2.x?

src/use-built-ins-plugin.js
hzoo: I'l use path.evaluate

src/use-built-ins-plugin.js
```diff
@@ -143,14 +143,27 @@ Please remove the call or use 'useBuiltIns: "entry"' instead.
...
+            has(definitions.instanceMethods, prop.value)
+          ) {
+            const builtIn = definitions.instanceMethods[prop.value];
+            warnOnInstanceMethod(state, `${obj.name}['${prop.value}']`);
+            addUnsupported(path, state.opts.polyfills, builtIn, this.builtIns);
+          } else {
+            const res = path.get("property").evaluate();
+            if (res.confident) {
+              const builtIn = definitions.instanceMethods[res.value];
+              warnOnInstanceMethod(state, `${obj.name}['${prop.value}']`);
```
mikesherov: should this be `res.value`?

hzoo: yeah forgot to change 👍

src/use-built-ins-plugin.js
```diff
@@ -90,8 +90,8 @@ to the "transform-polyfill-require" plugin
           if (isRequire(bodyPath)) {
             console.warn(
               `
-When setting 'useBuiltIns: true', polyfills are automatically imported when needed.
-Please remove the call or use 'useBuiltIns: "entry"' instead.
+When setting "useBuiltIns: true", polyfills are automatically imported when needed.
+Please remove the "import 'babel-polyfill'" call or use "useBuiltIns: 'entry'" instead.
```
existentialism: Should be "require('babel-polyfill')"

hzoo: lol almost forgot why it was 2 messages in the first place then

README.md
```diff
@@ -176,21 +174,60 @@ This option is useful for "blacklisting" a transform like `transform-regenerator
 
 ### `useBuiltIns`
 
-`boolean`, defaults to `false`.
+`boolean`, defaults to `true`.
```
yavorsky: maybe consider with `entry` option?

hzoo: what do you mean? I have #### `useBuiltIns: 'entry'` below

yavorsky: entry is a string :)

existentialism:
👏

PR #255, opened 2017-04-07
Update smoke-test

existentialism:


scripts/smoke-test.js
```diff
@@ -13,7 +13,7 @@ try {
   execSync("npm pack");
 
   console.log("Setting up smoke test");
-  fs.ensureDir(tempFolderPath);
+  fs.ensureDirSync(tempFolderPath);
```
existentialism: Probably the cause of all the intermittent failures

PR #264, opened 2017-04-10
Add forceAllTransforms option and deprecate Uglify target

### LINKED ISSUE #234: [2.0] Rename uglify option to useSyntax
similar to `useBuiltIns: true` we can have a `useSyntax` option.

or rename both to `builtins` and `syntax` or `useBuiltIns` and `compileSyntax`? (default to true)

It's not really related to uglify perse but just because uglify doesn't handle es6.

existentialism:
Fixes #234.

Added as a deprecation, but can straight remove if we want?

Also, still unsure about the name (`useTargetSyntax`?)?


yavorsky:
👍

Also not sure about `useSyntax` naming. `useTargetSyntax` maybe can be confusing (we need to specifify targetSyntax or `targets` are syntax targets).

Maybe focus on the code transformation? `useTransformations`/`useTransform`?

existentialism:
@yavorsky thoughts on `forceAllTranforms`?

With a `.babelrc.js` you would have:

```js
module.exports = {
  presets: [
    ["env", {
      targets: {
        chrome: 59,
        edge: 13,
        firefox: 50,
      },
      forceAllTransforms: process.env === "production"
    }],
  ],
};
```

yavorsky:
forceAllTransforms looks better. 👍
But on the other hand, `useBuiltIns: "entry"` make some transforms also. 🤔

existentialism:
@yavorsky @hzoo updated the name to `forceAllTransforms`... can land and rename it later?

nodkz:
@existentialism @yavorsky can you add a prefix to your deprecation messages?

I update dependencies, start a new production build and got such error:

screen shot 2017-08-10 at 17 33 02

Firstly I think that problem in new Webpack 3.5, but its and UglifyPlugin source code does not contain such massage. And look in my `.babelrc.js` and found `targets.uglify: true` for production build there.

Inside you use `console.log`, maybe you add wrapper over your messages `deprecate`, `info`, `error` etc?
Thanks.

existentialism:
@nodkz :+1:

PR #275, opened 2017-04-12
Add a reminder about include/exclude options [skip ci]

existentialism:
Hopefully fix some confusion reported in Slack?


yavorsky:
Maybe also add something like `If you want to include plugins not related to this preset, just add them to the `plugins` field directly.`?

PR #276, opened 2017-04-12
Better debug for use-built-ins-plugin.

yavorsky:
Enable `warnOnInstanceMethod` and fix its output.

Also, we could use `buildCodeFrameError` for this things in the future like we are doing for `check-const` fo ex.

### DIFF
--- a/src/use-built-ins-plugin.js
+++ b/src/use-built-ins-plugin.js
@@ -4,11 +4,11 @@ function isPolyfillSource(value) {
   return value === "babel-polyfill";
 }
 
-function warnOnInstanceMethod() {
-  // state.opts.debug &&
-  //   console.warn(
-  //     `Adding a polyfill: An instance method may have been used: ${details}`,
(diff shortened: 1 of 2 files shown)


src/use-built-ins-plugin.js
```diff
@@ -4,11 +4,11 @@ function isPolyfillSource(value) {
   return value === "babel-polyfill";
 }
 
-function warnOnInstanceMethod() {
-  // state.opts.debug &&
-  //   console.warn(
-  //     `Adding a polyfill: An instance method may have been used: ${details}`,
-  //   );
+function warnOnInstanceMethod(state, details) {
+  state.opts.debug &&
```
hzoo: Right now there's too many warnings being created when testing so not sure when to add this in.

yavorsky: @hzoo maybe just combine all instance warns into 1 message?

Issue #277, opened 2017-04-12
Reduce package size


evilebottnawi:
Package contains unnecessary directories and files - `coverage`, `.eslintrc` and maybe others. I can do PR if your accept this issue.

hzoo:
Sure. just add coverage/.eslintrc to https://github.com/babel/babel-preset-env/blob/master/.npmignore. The only other one would be contributing.md

hzoo:
Closed in https://github.com/babel/babel-preset-env/issues/281

PR #281, opened 2017-04-12
Chore: reduce package size.

evilebottnawi:
Ref: https://github.com/babel/babel-preset-env/issues/277


existentialism:
Should we just move to `files` key with `data` and `lib`?

hzoo:
Thanks for the issue/PR! 👍

PR #282, opened 2017-04-13
Clarify note about loading polyfills only once

darahak:
Reading the note, I wasn't sure if it could cause errors or not, and I had to look around for a little while.
Now it will say explicitly that loading `babel-polyfill` more than once will throw an error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,7 @@ npm install babel-polyfill --save
 This option enables a new plugin that replaces the statement `import "babel-polyfill"` or `require("babel-polyfill")` with individual requires for `babel-polyfill` based on environment.
 
 > NOTE: Only use `require("babel-polyfill");` once in your whole app. One option is to create a single entry file that only contains the require statement.
+> If `babel-polyfill` is loaded more than once, it will throw an error because these global polyfills can collide and provoke issues that are hard to trace.
 
 **In**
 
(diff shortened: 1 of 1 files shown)


hzoo:
I did put

>  Only use `require("babel-polyfill");` once in your whole app

But if this makes it clearer, we can do that. I can remove multiple imports if it's in the same file but if you actually import polyfill multiple times then yes it will error. Actually now that I think about it, using useBuiltIns won't `import babel-polyfill` so that error message won't be hit

existentialism:
Yeah, I'd prefer if we just combine the statements, something like:

> NOTE: Only use require("babel-polyfill"); once in your whole app. Multiple imports or requires of `babel-polyfill` will throw an error. We recommend creating a single entry file that only contains the require statement.

darahak:
@existentialism Your proposal has better phrasing. I would like to keep the rationale though.

existentialism:
Sure:

> NOTE: Only use require("babel-polyfill"); once in your whole app. Multiple imports or requires of babel-polyfill will throw an error since it can cause global collisions and other issues that are hard to trace. We recommend creating a single entry file that only contains the require statement.

darahak:
Thanks.
I updated this with your suggestion.

existentialism:
@darahak thanks!

PR #288, opened 2017-04-14
Update README with 'usage'/'entry'/false values.

yavorsky:
I understand that it's not the final API, but will be useful for those, who are trying alpha.


hzoo:
👍 forgot to do this, good catch

PR #295, opened 2017-04-17
Add failing test for for..of combined with Set object

JakeChampion:
It looks to be including the `web.dom.iterable.js` polyfill set, which is not required in this instance as no DOM apis are being called.

I think it needs to include `es6.symbol` as that is not included in the `es6.set` polyfill if I've understood the core.js source code correctly.


hzoo:
`new Set([1, 2, 3])[Symbol.iterator]()` matches https://github.com/babel/babel-preset-env/blob/28e0ce22c818940c7045e8bb7e99c677dd0c0122/src/use-built-ins-plugin.js#L109-L126, we would need to ignore if `arr` is a known built-in like Set.

hzoo:
Thanks for the PR! 

PR #312, opened 2017-04-28
Bump browserslist to 2.1.

yavorsky:
Use `browserslist` 2.0.
No changes that could affect us while we have [this condition](https://github.com/babel/babel-preset-env/blob/2.0/src/targets-parser.js#L91-L93).
Just size minification + parsing updates.

Maybe even it's worth to change base branch to `master` ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "babel-plugin-transform-es2015-unicode-regex": "7.0.0-alpha.9",
     "babel-plugin-transform-exponentiation-operator": "7.0.0-alpha.9",
     "babel-plugin-transform-regenerator": "7.0.0-alpha.9",
-    "browserslist": "^1.4.0",
+    "browserslist": "^2.1.2",
     "invariant": "^2.2.2",
     "semver": "^5.3.0"
   },
--- a/yarn.lock
+++ b/yarn.lock
@@ -1691,6 +1691,13 @@ browserslist@^1.4.0:
     caniuse-db "^1.0.30000639"
     electron-to-chromium "^1.2.7"
 
+browserslist@^2.1.2:
+  version "2.1.2"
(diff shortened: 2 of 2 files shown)


existentialism:
Hesitant to move this to straight over on master since it seems "last 2 versions" meaning has changed (probably not a big deal in practice tho)?

Also looks like autoprefixer is still on `^1.7` as well.

corbinu:
Autoprefixer now uses 2.0 their reasons for changing "last 2 versions" are detailed here: https://evilmartians.com/chronicles/autoprefixer-7-browserslist-2-released

duzun:
+1

existentialism:
@corbinu yep, and again, totally down to merge this into 2.x but we were discussing whether this should land on 1.x

corbinu:
@existentialism yes and you seems to be wondering how large of a deal the last two version changes were in that discussion. Hence my post to the article explaining the changes in detail, but don't worry I try to contribute to a discussion on babel again

existentialism:
@corbinu thanks for the link, sorry you seem to have gotten offended... was not the intention. Are you wanting it to land in 1.x?

corbinu:
@existentialism Honestly I have no preference. I see the arguments for both I was just trying to provide some helpful information to help in those closer to the project make a decision. I know that if major bumps of dependancies should be considered a breaking change or not is a hotly debated part of semver and usually taken in a case by case basis.

existentialism:
@corbinu def agree on the hotly contested part :) I think we'll probably land this in 2.x. Your link was indeed helpful, and we definitely welcome the contribution to the discussion.

Kovensky:
The change to `last n versions` should not affect us since we have https://github.com/babel/babel-preset-env/blob/20ddb722/src/index.js#L49 to filter out the browsers we have no data about.

existentialism:
Landed in https://github.com/babel/babel-preset-env/pull/319

PR #324, opened 2017-05-15
Backport: use preset-env and remove flow-strip-types

yavorsky:
Nothing special, just use `preset-env` for backport and remove `flow-strip-types` plugin.


hzoo:
Is this necessary to backport though?

yavorsky:
@hzoo For now it's still the main branch and few people asked me why we are using es2015 instead of env 😕

package.json
```diff
@@ -76,12 +76,7 @@
   },
   "babel": {
     "presets": [
-      [
-        "es2015",
-        {
-          "loose": true
-        }
-      ]
+      ["env"]
```
tleunen: Isn't it better to use the current code instead of always the previous version?
So replacing `["env"]` with `"./src"`?

hzoo: If you have a bug then it won't build itself

hzoo:
@yavorsky lol really..

PR #342, opened 2017-05-31
add maintained node target Fixes #341

### LINKED ISSUE #341: option for node: 'maintained' that enables rules for the oldest maintained node version.
See this chart for maintenance cycle: https://github.com/nodejs/LTS

graingert:
these flags suggest that users with babel-register should use 'current' and `babel src dist` should use 'maintained'

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,9 +110,10 @@ The [data](https://github.com/babel/babel-preset-env/blob/master/data/plugins.js
 
 ### `targets.node`
 
-`number | string | "current" | true`
+`number | string | "current" | "maintained" | true`
 
 If you want to compile against the current node version, you can specify `"node": true` or `"node": "current"`, which would be the same as `"node": parseFloat(process.versions.node)`.
+If you want to compile against the latest [maintained](https://github.com/nodejs/LTS) node version, you can specify: "node": "maintained".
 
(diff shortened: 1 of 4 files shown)


hzoo:
cc @mylesborins on thoughts on this option in general (and I guess hardcoding is ok?).

Not sure about this option because at some point (the day of the switch), it will just target the next version even if you yourself haven't updated? Unless we should assume that if you pick this option it means that you intend to upgrade beforehand?

graingert:
@hzoo the same problems exist with browser version ranges

graingert:
If we're going to go with hard-coding I'll make a build script that runs pulls the data out of https://github.com/nodejs/LTS/blob/master/schedule.json

niieani:
I think it would be better to call this option "lts" instead of "maintained". after all non-LTS are also _maintained_, just not _supported_. Another option would be that very word: "supported".

graingert:
@niieani I don't mind about the word

existentialism:
Let's go with `lts`.

> If we're going to go with hard-coding I'll make a build script that pulls the data out of https://github.com/nodejs/LTS/blob/master/schedule.json

👍

hzoo:
Thanks for the PR! 

PR #352, opened 2017-06-12
updated package.json

EsrefDurna:
updated package babel-plugin-transform-es2015-block-scoping from 6.23.0 to 6.24.1


hzoo:
Thanks for the PR, but our package.json uses `^` so you will get the update after a `npm install` so this isn't necessary!

Issue #373, opened 2017-07-13
Language spec targets instead of versions of browsers/Node.js


rtsao:
I'd love for the ability to do something like this:

```
{
  "presets": [
    "env", {
      "targets": {
        "languageSpecification": "ES2016"
      },
    }
  ] 
}
```

You might say this doesn't make sense, but this preset includes things beyond just `babel-preset-es2015`, so being able to use language version would be nice as this preset evolves.

XaveScor:
Why? Your target is browser or Nodejs. Not some standard.

rtsao:
Not if you're publishing a universal JS library. You don't know what the consuming browser/runtime is – you just make an arbitrary decision as to what JS spec your published module conforms to.

XaveScor:
I know the minimal browser/node version. I rely that the browser/node developer save the backward compatibility.
Ok, let's imagine we decide your proposal. Let's look ES5 support. http://kangax.github.io/compat-table/es5/ I want to support IE11 and Safari. What I need to do?
Your proposal is tangle for developers. After we need to look how browsers support every feature. Why? We can do that in the babel side.

milesj:
@rtsao So publish the source code alongside the compiled code and give the consumer the ability to compile the source code down themselves using their own build process and configuration.

rtsao:
> So publish the source code alongside the compiled code and give the consumer the ability to compile the source code down themselves using their own build process and configuration.

Personally, I think this approach is unscalable and shifts too much burden to downstream consumers of packages. For example, if I'm authoring my JS package in a language other than JavaScript, merely providing the source code may not be very helpful. It's common for package authors to employ highly complex, specialized code transformations and build steps (arbitrary syntax/macros, cross-compilation from other languages, etc). Requiring the consumer configure/control this process is probably not going to be a good experience even if a convention could be established.

> The community needs to start heading that direction instead of compiling down to ES3 or assuming what target they need.

[...]

PR #376, opened 2017-07-17
Normalize module format of plugins/built-ins data

rtsao:
Slight refactor of `data/plugin-features.js` that makes it match the same format of [`data/built-in-features.js` where the data is organized by ECMAScript year](https://github.com/babel/babel-preset-env/blob/db25efd99a1d58a8eb08abcf0b495166000872a9/data/built-in-features.js#L181).

Resolves https://github.com/babel/babel-preset-env/issues/373 by allowing package authors to import `babel-preset-env/data/plugin-features.js` and `babel-preset-env/data/plugin-features.js` and assemble the desired exclusions/inclusions (corresponding to a language target) on their own. There's no need for a first-class interface as long as the format of `data/plugin-features.js` and `data/built-in-features.js` is preserved (whereby a clean userland implementation is possible).

Note: this PR does not alter the output produced by the `build-data` script in any way. The output is byte-for-byte exactly as what is currently committed to master.

### DIFF
--- a/data/plugin-features.js
+++ b/data/plugin-features.js
@@ -1,5 +1,4 @@
-module.exports = {
-  // es2015
+const es2015 = {
   "check-es2015-constants": {
     features: [
       "const",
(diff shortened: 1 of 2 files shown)


hzoo:
I'm cool with the refactor and we probably won't change the file but we can't 100% guarantee that as a public "API"

rtsao:
Thanks, that's good enough for me. I think ideally https://github.com/babel/babel-preset-env/issues/373 is made part of the actual API, but since there was lot of pushback, this will suffice until a decision is made.

PR #380, opened 2017-07-22
Fixed "node: current" example

leggiero:
`parseFloat` is not applied anymore to "node: current" logic:
The parseFloat was wrong, as `parseFloat('6.3.2')` will parse as 6.3, but `parseFloat('6.10.2')` will output 6.1 that is not correct.

Ref: https://github.com/babel/babel-preset-env/blob/master/src/targets-parser.js#L73


yavorsky:
@leggiero Thanks!

PR #384, opened 2017-07-28
Add stage-3 plugin option

existentialism:
Adds support for enabling transformation of **shipping** stage 3 features.

**EDIT** Updating this to:

* Enable syntax-only if transform is not needed
* Update mapping for optional catch binding (https://github.com/kangax/compat-table/pull/1159)


chicoxyzzy:
> 🚲 🏠 the option name :)

I think `unspeccedProposals` may be a good name for it

data/plugins.json
```diff
@@ -238,5 +237,14 @@
     "ios": "10",
     "opera": "45",
     "electron": "1.7"
-  }
+  },
+  "transform-async-generator-functions": {},
```
existentialism: These will get filled out once we land https://github.com/babel/babel-preset-env/pull/402

src/index.js
```diff
@@ -7,13 +7,36 @@ import { defaultWebIncludes } from "./default-includes";
...
-import { prettifyTargets, semverify } from "./utils";
+import { filterStageFromList, prettifyTargets, semverify } from "./utils";
 import type { Plugin, Targets } from "./types";
 
+const getPlugin = (pluginName: string) => {
+  const plugin = availablePlugins[pluginName];
+
+  if (!plugin) {
+    throw new Error(
+      `Could not find plugin "${pluginName}". Ensure there is an entry in ./available-plugins.js for it.`,
```
existentialism: This kept tripping me up 😜

yavorsky: lol

existentialism:
I'll create a followup to update README, but we should def discuss if there's a better name for the option too.

PR #397, opened 2017-08-18
Fix README. rename Babili to babel-minify [skip ci]

mediba-takeda:


existentialism:
@mediba-takeda thanks!

PR #398, opened 2017-08-18
Add Node usage

corysimmons:
I can never remember the syntax and end up coming here to copy/paste this snippet all the time.

I assume other people want to as well, or at least make it clear for newbs that this can be used with Node without them having to scroll.


hzoo:
> without them having to scroll 😅 

Ok

PR #404, opened 2017-08-30
Add onPresetBuild callback support.

yavorsky:
[I've described](https://github.com/babel/website/pull/1356#issuecomment-325742273) that to get rid of babel-preset-env-standalone (we need to copy-paste index.js stuff every time we change it for the original repo) and being able to use unpckg or bundle.run we need to provide `onPresetBuild` support with used plugins/presets as arguments.

So, what we currently have:
config:
```
{
      targets: { ie: 11 },
      onPresetBuild({ 
        targets,
        transformations,
        transformationsWithTargets,
        polyfills,
        polyfillsWithTargets,
        usedPolyfillsInFiles,
        modules,
      }) {
      },
      useBuiltIns: "usage"
}
```

Also, it will be useful for `.babelrc.js` where we can detect/log/use plugins and polyfills was actually used.

Callback arguments:

**targets**: just list of the current prettified targets:
`{"chrome": "55"}`

**transformations**: filtered plugins according to the targets

**transformationsWithTargets**: the same, but each item with related plugins

**polyfills**/**polyfillsWithTargets** - the same as plugins, but useBuiltIns. w/o useBuiltIns just empty Sets (maybe change to `null` ?)

**usedPolyfillsInFiles**: for `useBuilIns: 'usage'`. `Array` with object according to each file with file path and `Array<transformationsWithTarget>` of polyfills used specifically in each file:

```
[
 {polyfills :[{ name: "es6.promise", targets: { ie: "11" }}], path: "src/in.js" },
 {polyfills :[], path: "src/in.js" }
]
```
**modules**:  just an module option ('amd', 'commonjs', etc)

cc @Daniel15 

Also, I've updated `debug-fixtures.js` to accept `options.js`.


Daniel15:
>  we need to provide onPresetBuild support with used plugins/presets as arguments.

So is the idea that `babel-preset-env` would have some way of returning a list of the plugins to use, and then we could use the versions of those plugins bundled with babel-standalone? We need to ensure that the standalone version of babel-preset-env does not bundle its own copy of the plugins (I think that's an issue with it today).

yavorsky:
@Daniel15 Yes, but we have 2 additional plugins there:
[use-built-ins-entry-plugin](https://github.com/babel/babel-preset-env/blob/2.0/src/use-built-ins-entry-plugin.js) and [use-built-ins-plugin](https://github.com/babel/babel-preset-env/blob/2.0/src/use-built-ins-plugin.js).

> We need to ensure that the standalone version of babel-preset-env does not bundle its own copy of the plugins and instead just reuses the babel-standalone ones

I think there we have 2 ways: [babel-env-standalone](https://github.com/babel/babel-standalone/pull/90) or accept plugins option where we can pass plugins from `Babel` during runtime.
Fresh ideas are welcome :)

Daniel15:
`babel-env-standalone` is likely fine. babel-standalone exposes all available plugins as `Babel.availablePlugins`:


We might be able to use some Webpack setting so that `require('babel-plugin-foo')` is rewritten to `Babel.availablePlugins.foo` (so it avoids bundling it)? We could register `use-built-ins-entry-plugin` and `use-built-ins-plugin` using `Babel.registerPlugins` so that it'd work for those too.

Daniel15:
I should say that I'm not a Webpack expert so any Webpack expert advice is appreciated 😛

yavorsky:
@Daniel15 Can you please publish `babel-env-standalone` to npm. Then we can try to use it with bundle.run or unpckg , etc.

yavorsky:
@Daniel15 I've tried to use [webpack-bundler-analizer](https://github.com/th0r/webpack-bundle-analyzer) to detect what packages are really large and this is the result:
The largest one is **caniuse-lite** which uses browserslist:
screen shot 2017-08-30 at 22 49 38

I think we can omit browserslist for standalone version and parse query by sending requests to external service. I'm thinking about http://browserl.ist, but seems like it doesn't support JSON requests, so I can send PR to implement this. What do you think?

Daniel15:
We can't always send requests to an external service in babel-standalone,
as we can't assume it's running in a browser. Some use cases are not in a
browser, for example as part of server side JS in environment other than
Node.js. That was actually the original reason for building
babel-standalone... I needed it for ReactJS.NET (which uses V8 from C#)
after jstransform was deprecated.

Additionally, some use cases may not have internet access (eg. Corporate
tools on machines with no external internet access).

Maybe we could bundle that stuff separately (so you'd load browserlist.js
in addition to env.js) and optionally allow people to use some external
service instead of loading the entire file.

I can publish the existing env-standalone package soon.

yavorsky:
@Daniel15 
> We can't always send requests to an external service in babel-standalone,
as we can't assume it's running in a browser

There rather I meant that we can have different standalone builds - *full* and *w/o plugins and browserslist + caniuse-lite*.

> Maybe we could bundle that stuff separately (so you'd load browserlist.js
in addition to env.js) and optionally allow people to use some external
service instead of loading the entire file.

I think you need browserslist in 90% of cases when you are using preset-env.
Today I've tried to reduce bundle size for babel-env-standalone and after removing unused regions the bundle size reduced by ~65%. So there is a good direction.

About this PR: it wasn't intended to reduce standalone version's size. It just possibility to detect which polyfills/plugins were used and to output them like we are doing for REPL. So we can include preset-env 2.0 to the new REPL. To prevent copy-pase of the main file between preset-env and env-standalone and adding own onPresetBuild wrapper manually this PR will be usefull also.
To reduce bundle size we need to remove some unused dependencies as I mention above.

Daniel15:
> @Daniel15 Can you please publish babel-env-standalone to npm? Then we can try to use it with bundle.run or unpckg , etc.

Sorry for the delay in replying to this. I just published the latest version that's in the babel-standalone repo.

hzoo:
Thanks for the PR! 

PR #409, opened 2017-09-06
add basic example [skip ci]

hzoo:
Via https://twitter.com/suchipi/status/905278280303079425

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,16 @@
 npm install babel-preset-env --save-dev
 ```
 
+Acts as a replacement for preset-es2015,16,17,latest
+
+```json
+{
+  "presets": ["env"]
+}
+```
+
+You can target specific versions with browserslist
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,6 +6,16 @@
 npm install babel-preset-env --save-dev
 ```
 
+Acts as a replacement for preset-es2015,16,17,latest
```
suchipi: Maybe, "Without any configuration options, it will include everything you would get from babel-preset-es2015, babel-preset-es2016, babel-preset-es2017, and babel-preset-latest". It's good to be clear and explicit about these things (especially for someone like me who does a ctrl+f for "I'm coming from a deprecation warning on babel-preset-latest, how do I get it to just work"

suchipi: Darn, I was too slow 😉

PR #416, opened 2017-09-16
Docs: fix claim about node versions in readme

bakkot:
Arrow functions [will in fact](https://github.com/babel/babel-preset-env/blob/be7a61c368149cd7e9d4202ecec6f71c56f2eb77/data/plugins.json#L17) be transformed when building for Node 4.


yavorsky:
@bakkot Thanks!

PR #441, opened 2017-10-13
[1.x]  Add polyfills for ES6 static Object methods

danez:
These functions do exist in ES5, but we still need to load the polyfills in ES5 browsers (like IE11), as the functions work sligthly different in ES5 vs ES6

We noticed this when trying to basically run the following code which throws in IE11 but works in latest Chrome:

```js
Object.keys(123);
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys#Notes

Can also port to 2.0 if this is okay.

//cc @sevinf


existentialism:
@danez let's port this to babel/babel (the new 2.x) as well

danez:
Oh it was moved already? Didn't know about that. Will do it!

hzoo:
@danez we moved the code but i failed at moving issues correctly 😛 (need to move open issues + some closed issues, I only moved older issues without a label)

PR #458, opened 2018-05-09
Update browserslist to the latest

devongovett:
Parcel users are getting "Unknown version 65 of chrome" with the old version since we depend on the latest browserslist already so there are 2 copies installed. See issue https://github.com/parcel-bundler/parcel/issues/1036

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "babel-plugin-transform-es2015-unicode-regex": "^6.22.0",
     "babel-plugin-transform-exponentiation-operator": "^6.22.0",
     "babel-plugin-transform-regenerator": "^6.22.0",
-    "browserslist": "^2.1.2",
+    "browserslist": "^3.2.6",
     "invariant": "^2.2.2",
     "semver": "^5.3.0"
   },


loganfsmyth:
What were the breaking changes that prompted it to publish a `3.x` and do any of them affect `preset-env`?

devongovett:
looks like maybe just the default query changed? https://github.com/browserslist/browserslist/blob/master/CHANGELOG.md#30

existentialism:
IIRC 2.x -> 3.x changed the `default` query from `> 1%` to `> 0.5%` (which doesn't matter much because preset-env [doesn't currently support](https://github.com/babel/babel/issues/7405) the `default` query), and removed country statistics.

loganfsmyth:
Cool, then pretend I'm not even here :P

hzoo:
Ok I unarchived the repo for the 1.x change. I don't know what to think about browserslist because everytime we update the version the `last 2 versions` change anyway so it's like ok.. breaking change anytime? Thanks @devongovett 

Also to others: this doesn't mean we will back port features, etc - just fixes and usually only if it's really breaking people.

hzoo:
Would like to know if there's a way for babel to just read the root browserslist rather than the one in the dep then.

existentialism:
@devongovett released as 1.7, thanks for the PR!

